PR #4, opened 2018-12-03
Refactured snap.py

mathiasbockwoldt:
This is a very extensive pull request. It is much longer than it should be, but I hope it is ok, since we are still in the beginning, when much is to be tested. A short changelog of the more important changes. The changelog is also in the commit message of the main commit.

Changed top to ps. Following reasons:
- ps is independent on locale (top turns e.g. CPU 2.3% to 2,3% depending on locale)
- ps does not produce the header that has to be shaved of
- ps is (arguably) more adaptable/configurable

Rewrote extract_processes and its test accordingly, so it works with the new ps solution.

Renamed get_slurm_project to get_slurm_projects (with s) as it will get all user-to-project mappings on the node.

Implemented get_slurm_projects. It will only get the active users (and their projects) on the current cluster, since the snap script will run on each node and does not need to know about other nodes.

Managed ignored users in another way that should (in my opinion) work better in future with config files.

I split take_snapshot into two functions, one to gather all data and return a list of lists, and another function that writes everything to file or stdout. This way, the first function is more easily testable and the second function could easily be changed to a database or whatever without tinkering with the first. Naming of the two functions may be unfortunate/unclear.

The second function uses a special wrapper function that enables easy writing to stdout or file.

Memory is saved in MiB, since the percentage is dependent on the node.

Slurm-specific stuff was tested on Stallo :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 __pycache__/
 venv/
 *.egg-info/
+.pytest_cache/
+
--- a/sonar/snap.py
+++ b/sonar/snap.py
@@ -1,4 +1,4 @@
-#/usr/bin/env python3
+#!/usr/bin/env python3
 
 import sys
 import os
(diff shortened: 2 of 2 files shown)


mathiasbockwoldt:
The main commit is much too large! I will try to keep my commits more concise in future.
I also had some problems with the setup script that I couldn't fix, so I added a good old `if __name__ == '__main__'` to the snap code to be able to test the script.

bast:
Thanks! Excellent changes. I think it would be great if you could document in the README with a sentence or two why we chose `ps` and not `top`. The explanation you gave is very good but I am almost sure I will forget it and in 5 years somebody else will think; "hmm... why did they use `ps` and not `top`? Maybe I should change that ..."

PR #7, opened 2018-12-07
Map

mathiasbockwoldt:
First draft of the mapping module (module 2). It certainly does not do everything we anticipated, but some processing is done. The example mapping file is currently for "normal" computers (having Firefox and Skype) and not for HPC.


bast:
Thanks! Looks good. I have suggestions for some minor changes but I will submit them as PR for later review.

PR #12, opened 2018-12-07
Refactor of the mapping.

bast:
Significant changes:
- a bit less code
- implemented a cache for the lookup function

### DIFF
--- a/sonar/cli.py
+++ b/sonar/cli.py
@@ -33,8 +33,8 @@ def main():
     parser_map = subparsers.add_parser('map', help='Parse the system snapshots and map applications. Run this only once centrally. Supposed to run e.g. once a day.')
     parser_map.add_argument('--snap-dir', default='', help='Path to the directory with the results of sonar snap. If empty, the current directory will be assumed.')
     parser_map.add_argument('--output-file', default='', help='Output file. Leave empty or provide - for stdout.')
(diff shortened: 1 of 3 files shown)


bast:
OK - I will remove the f-string in this patch.

mathiasbockwoldt:
In `map.py` lines 26-33, it would remove an indentation level to use something like:
```python
            try:
                with open(file_name) as f:
                    f_reader = csv.reader(f, delimiter='\t', quotechar='"')
                    for k, v in f_reader:
                        l.append((k, v))
            except FileNotFoundError:
                sys.stderr.write(f'ERROR: file {file_name} not found\n')
            except TypeError: # this will catch file_name is None
                pass
```
(without the `if file_name is not None`).

Also, I think that using `sys.stderr.write` is maybe not as elegant as `print(..., file=sys.stderr)` or even using the logging module.

Both changes are of course kind of personal taste, though.

bast:
I agree about `sys.stderr.write`.

I don't like except ... pass constructs (unless there is a very good motivation for them) and I think the `if file_name is not None` is more explicit.

bast:
... and the reason why I don't like them is that I think they allow errors that we haven't imagined at this point to fall through the cracks.

bast:
I changed to print. I agree that logging would be even better but I postponed that for later :-)

mathiasbockwoldt:
Thank you! Yes, logging can be postponed so far :) The `except: pass` objection is good :)
Nice caching!

PR #13, opened 2018-12-10
 Added support for config files.

mathiasbockwoldt:
Precendence:
  command line arguments > config file > hard-coded default values
Until now, not all config file parameters can be changed with command line arguments but that should be easy to add.

### DIFF
--- a/example_config.conf
+++ b/example_config.conf
@@ -1,0 +1,20 @@
+# Command line arguments SHOULD (FIXME :) ) take precedence over values in this config
+
+[DEFAULT]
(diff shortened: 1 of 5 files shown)


mathiasbockwoldt:
We might consider to pass the whole config dictionary to the respective main functions of map and snap instead of passing every argument explicitly.

bast:
Thanks! Very good - merging. I agree that passing a config dictionary might be good but it should perhaps not traverse the entire code but as you say for the main functions it probably makes things cleaner.

PR #25, opened 2019-01-28
Added hack to allow snap and map to run without Flask being installed

mathiasbockwoldt:


bast:
Thanks! This will be a good reminder for me to keep them more decoupled.

PR #45, opened 2023-06-16
Refactor ps

lars-t-hansen:
Refactors the code for `ps` by breaking out utilities, the code that runs and parses the `ps` command, and the code that runs and parses the `nvidia-smi` command.  No functional changes.

PR #46, opened 2023-06-16
Refactor jobs

lars-t-hansen:
This creates a JobManager abstraction to remove the dependency on SLURM.  This patch sits on top of #45.


bast:
I will rebase it to make it easier to see what is added in this PR.

bast:
I need to study the code a bit more to fully understand but I am sure it is better than before so let's merge. It is a good idea to abstract the Slurm part away.

PR #48, opened 2023-06-16
new command 'Read' which fetches data from specified filed

benteb:
Didn't want to interfere with the Analyze command and read_users function, so I made the (temporary) new command 'Read'.  It takes arguments data_dir and data_field, where data_dir is the directory of csv files and data_field is an integer indicating what field to read (0=time stamp, 1=hostname, 2 = corse on node, etc.)  

```$ sonar read --data-dir=2023 --data-field=5 ``` will for example fetch the data_field=5 (correspond to processes) from all readable files in 2023 and subdirectories. Will fail if it encounters unreadable files like compressed files. 

New functions are in read_fields.rs. I made the functions general to replace read_users and open up other fields. The function only accepts directories as argument and not single files as I didn't see an immediate reason to analyze single files.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-use clap::{ Parser, Subcommand };
+use clap::{Parser, Subcommand};
 use std::collections::HashSet;
 use std::fs;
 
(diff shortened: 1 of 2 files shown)


bast:
Thanks! This is very good step towards analyze. We can merge or keep on the side-branch, depending on your preference.

bast:
Closing this in favor of Jobanalyzer as we have agreed elsewhere.

Issue #44, opened 2023-06-16
Add a pre-check whether host/node is GPU-aware before running any other GPU commands


bast:
Follow-up from #40. This pre-check should not be really needed since the subprocess is implemented to hopefully never crash or hang but I would still feel better if we do a quick check first before running other commands.

lars-t-hansen:
For nvidia we could maybe:

- check for /dev/nvidia*
- check to see if nvidia-smi exists
- try to run nvidia-smi -L to see if there's any output at all

I'd love for there to be something absolutely solid but those are my best ideas so far, for run-time detection.  There's `module avail` but then we take a dependency on `module` plus I know for a fact we have non-NVIDIA machines that have the NVIDIA stack.

Other options are:

- command line switch to `sonar ps` to tell it to look for nvidia cards, but this adds friction to setup
- compile-time switch, but this adds friction to deployment

lars-t-hansen:
There's some variant of `lspci | egrep 'VGA.*(NVIDIA Corporation|Advanced Micro Devices)'` I guess.  On my systems this finds all the cards.  Indeed, finding the NVIDIA string means we should be able to run nvidia-smi, finding the AMD string that we should be able to run rocm-smi.

lars-t-hansen:
A complementary approach is that used by [the sysinfo prototype](https://github.com/NAICNO/Jobanalyzer/tree/main/code/sysinfo): it uses Go's `LookPath` primitive to resolve the paths for `nvidia-smi` and `rocm-smi` to ensure that the programs are found in the path but not in the current directory.  This doesn't quite address the core of the problem - we want to be sure that the probe programs don't hang - but may in any case be desirable.

lars-t-hansen:
As #188 shows, the presence of nvidia-smi is by itself insufficient.

PR #49, opened 2023-06-19
Amd gpu support

lars-t-hansen:
This parses the output of rocm-smi to generate GPU usage data for AMD, as for NVIDIA.  It's a bit of a hack but is reliable in my testing - I have a machine with two NVIDIA cards, and some test code that uses both of them.

I think that our nervousness about running nvidia-smi (see #44) is going to be even more acute for rocm-smi.

~~(The base rev here is not the tip of my work queue but the underlying main refactoring of ps.rs, #45, a little rebasing work will be required to land this if all the other patches land first.)~~

### DIFF
--- a/src/amd.rs
+++ b/src/amd.rs
@@ -1,0 +1,303 @@
+// Get info about AMD graphics cards by parsing the output of rocm-smi.
+// This is pretty hacky!  Something better than this is likely needed and hopefully possible.
+//
+// Note that the returned information is keyed by (device, pid) so that if a process uses multiple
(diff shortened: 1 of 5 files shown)


bast:
When I try to rebase this on top of `main`, then `src/ps.rs` conflicts. Could you please have a look? Or we can do that together. Or I try it :-)

lars-t-hansen:
I'll have a look.

lars-t-hansen:
Rebased patch tested on ML4 with AMD cards, seems to work OK.

bast:
Thanks a lot!

PR #54, opened 2023-06-21
Propagate subprocess error

### LINKED ISSUE #53: Propagate errors from the command submodule
See #51: if running a command fails, we want to know about this.

lars-t-hansen:
Fix #53.  This sits on top of the rebased AMD patch.


lars-t-hansen:
Tests fail because the integration environment does not have traceroute, which is a command I used to trigger a timeout.  Will look for something else.

lars-t-hansen:
I've concluded I really don't understand how patch stacks are supposed to work with git, I'll close this PR and resubmit.

PR #55, opened 2023-06-22
Handle large inputs

### LINKED ISSUE #51: Large output is a reality, and killing the subprocess is not a workaround
Basically, running ps on "big enough" systems or asking for the unfiltered ps output will create enough output to make the pipe fill up and the process hang, in turn causing the subprocess to be killed by the timeout in command.rs: https://github.com/NordicHPC/sonar/pull/50#issuecomment-1600497721.

There are really two bugs here.  One is that if the subprocess is killed there should be some report up the call chain so that whoever started the process can take evasive action, log an error, etc.  (In general there are probably good reasons why we should try to catch errors high up and log them.)  The other is that we need to fix the hangs caused by too-large output.

lars-t-hansen:
This fixes #51 by restructuring safe_command a bit, the changes are local to that file.

### DIFF
--- a/src/amd.rs
+++ b/src/amd.rs
@@ -16,26 +16,32 @@
 // about the usage of various processes on the various devices.  We divide the utilization of a
 // device by the number of processes on the device.  This is approximate.
 
-use crate::command;
+use crate::command::{self, CmdError};
 use crate::nvidia;
 #[cfg(test)]
 use crate::util::map;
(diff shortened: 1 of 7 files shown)


lars-t-hansen:
I've concluded I really don't understand how patch stacks are supposed to work with git, I'll close this PR and resubmit.

PR #66, opened 2023-07-29
make output fields self-identifying; closes #41

### LINKED ISSUE #41: Make output fields self-identifying?
As we experiment with sonar and add more fields and move it to new systems there will come a time when some log data are older and some are newer and data of different ages will have different columns.  For a while it will be good enough to always add new columns at the end of the line, but in the long term this will likely become brittle.  It may be good to add an identifier to each field to reduce the impact of this problem.  The identifier can be a short name and can prefix the field, for example, 
```
date=...,host=...,cores=...,user=...,job=...,cmd=...,cpu%=...,kib=...,gpus=...,gpu%=...,gpumem%=...,gpukib=...
```
would do it for now.  Names (columns) could be removed if they turn out to be worthless but names would never be reused with a different meaning.

bast:
Summary:
- changes output format (which is explained in README.md)
- bumps version to 0.7.0

Please see the change in README.md for how the columns are defined. @lars-t-hansen can you please browse through this change carefully?

I am unsure about the `gpus` part.
- If a process would use GPUs 1, 3, and 7: `"gpus={1, 3, 7}"`
- If a process would use no or unknown GPUs: `gpus={}`

Do we want to distinguish no GPUs and unknown GPUs? If yes, how?

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,22 +86,45 @@ Here is an example output:
 ```console
 $ sonar ps
 
-2023-07-27T18:29:17+02:00,somehost,8,user,,alacritty,3.7,214932
-2023-07-27T18:29:17+02:00,somehost,8,user,,slack,2.4,1328412
-2023-07-27T18:29:17+02:00,somehost,8,user,,X,0.8,173148
-2023-07-27T18:29:17+02:00,somehost,8,user,,brave,15.5,7085968
-2023-07-27T18:29:17+02:00,somehost,8,user,,.zoom,37.8,1722564
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=.vim-wrapped,cpu%=1.9,cpukib=7228,gpus={},gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=node,cpu%=1.8,cpukib=79332,gpus={},gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=slack,cpu%=0.7,cpukib=591720,gpus={},gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=X,cpu%=1.5,cpukib=224416,gpus={},gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=brave,cpu%=12.1,cpukib=3075300,gpus={},gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=alacritty,cpu%=1.2,cpukib=286196,gpus={},gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=sonar,cpu%=9,cpukib=372,gpus={},gpu%=0,gpumem%=0,gpukib=0
 ```
 
 The columns are:
-- time stamp
-- hostname
-- number of cores on this node
-- user
-- Slurm job ID (empty if not applicable)
-- process
-- CPU percentage (as they come out of `ps`)
-- memory used in KiB
+- `v`: version (in the format n.m.o, following semantic versioning)
+- `time`: local time stamp (in ISO time without fractional seconds but with TZO)
+- `host`: host name (FQDN)
+- `cores`: number of cores on this node (positive integer)
+- `user` : username owning the process/command (it can also be "unknown" and "zombie")
+- `job`: job ID (positive integer; 0 if not applicable)
+- `cmd`: process/command
+- `cpu%`: CPU percentage (in percent of one core; as they come out of `ps`)
+- `cpukib`: CPU memory used in KiB
+- `gpus`: GPU devices (the card indices are 1-based; more about it below)
+- `gpu%`: GPU percentage (sim across cards)
+- `gpumem%`: GPU memory percentage (in percent of memory across all cards)
+- `gpukib`: GPU memory used in KiB (sum across cards)
+
+`gpumem%` vs `gpukib`:
+The difference is that on some cards some of the time it is possible to
+determine one of these but not the other, and vice versa. For example, on the
+NVIDIA cards we can read both quantities for running processes but only
+`gpukib` for some zombies. Since we can detect the total amount of memory here
+we could translate `gpukib` into `gpumem%`, though. On the other hand, on our
+AMD cards there is no support for detecting the absolute amount of memory used,
+nor the total amount of memory on the cards, only the percentage of gpu memory
+used. Rather than encoding the logic for dealing with this, it seemed better
+for the time being to report what we can report and let the analyzer sort it
+out.
+
+`gpus` are GPU devices:
+- If a process would use GPUs 1, 3, and 7: `"gpus={1, 3, 7}"`
+- If a process would use no or unknown GPUs: `gpus={}`
 
 
 ## Collect results with `sonar analyze` :construction:
--- a/README.md
+++ b/README.md
@@ -86,13 +86,13 @@ Here is an example output:
 ```console
 $ sonar ps
 
-v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=.vim-wrapped,cpu%=1.9,cpukib=7228,gpus={},gpu%=0,gpumem%=0,gpukib=0
-v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=node,cpu%=1.8,cpukib=79332,gpus={},gpu%=0,gpumem%=0,gpukib=0
-v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=slack,cpu%=0.7,cpukib=591720,gpus={},gpu%=0,gpumem%=0,gpukib=0
-v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=X,cpu%=1.5,cpukib=224416,gpus={},gpu%=0,gpumem%=0,gpukib=0
-v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=brave,cpu%=12.1,cpukib=3075300,gpus={},gpu%=0,gpumem%=0,gpukib=0
-v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=alacritty,cpu%=1.2,cpukib=286196,gpus={},gpu%=0,gpumem%=0,gpukib=0
-v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=sonar,cpu%=9,cpukib=372,gpus={},gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=.vim-wrapped,cpu%=1.9,cpukib=7228,gpus=none,gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=node,cpu%=1.8,cpukib=79332,gpus=none,gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=slack,cpu%=0.7,cpukib=591720,gpus=none,gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=X,cpu%=1.5,cpukib=224416,gpus=none,gpu%=0,gpumem%=0,gpukib=0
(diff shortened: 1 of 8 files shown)


bast:
Also if you could test the branch on the GPUs you have access to, that would help a lot. I haven't tested this outside of CPUs. But the tests are preserved.

README.md
```diff
@@ -86,22 +86,45 @@ Here is an example output:
...
+- `job`: job ID (0 if not applicable)
+- `cmd`: process/command
+- `cpu%`: CPU percentage (as they come out of `ps`)
+- `cpukib`: CPU memory used in KiB
+- `gpus`: GPU devices (more about it below)
+- `gpu%`: GPU percentage
+- `gpumem%`: GPU memory percentage
+- `gpukib`: GPU memory used in KiB
+
+`gpumem%` vs `gpumem%`:
```
lars-t-hansen, line 113: This should probably read "`gpumem%` vs `gpukib`".

README.md
```diff
@@ -86,22 +86,45 @@ Here is an example output:
...
+we could translate `gpukib` into `gpumem%`, though. On the other hand, on our
+AMD cards there is no support for detecting the absolute amount of memory used,
+nor the total amount of memory on the cards, only the percentage of gpu memory
+used. Rather than encoding the logic for dealing with this, it seemed better
+for the time being to report what we can report and let the analyzer sort it
+out.
+
+`gpus` are GPU devices:
+- If a process would use GPUs 1, 3, and 7: `"gpus={1, 3, 7}"`
+- If a process would use no or unknown GPUs: `gpus={}`
```
lars-t-hansen, line 127: There's a difference between "no gpus" and "unknown gpus".  The reason for this is that in the code for nvidia cards, most information is grabbed using `nvidia-smi pmon`, and this gives us precise information, but it does not report everything, in particular, it does not report defunct or zombie processes that hold onto GPU memory.  For that we use `nvidia-smi --query-compute-apps`, but that command does not give us information about the cards being used, and so if there are any of those we use the "unknown" tag.

While it would be nice to clean that up by using the programmatic interface to the nvidia cards (and I really should do that...), I think it is useful to preserve the distinction between "no gpus" and "unknown gpus" here, it allows us to express more.

I don't understand why the braces are desirable in the syntax here.  From a parsing point of view, the field value "gpus=1,3,7" seems almost ideal because it's just a string split operation in the ingestor.  And with that background, we could use eg "gpus=none" for the "no gpu" case (though see below) and "gpus=unknown" for the unknown case, it's easy to distinguish those two cases from the list-of-card-numbers case.

On that topic, a useful optimization for the log writer is the use of default values that are agreed between the logger and the ingestor.  The default value for `gpus` would be `none` and sensibly the field does not appear in the output at all.  For several other fields, a sensible default is zero and if the value is zero we could (optionally of course) omit the field.  The ingestor will have to deal with missing fields and obsoleted fields and newly-introduced fields anyway, so this adds very little complexity.

README.md
```diff
@@ -86,22 +86,45 @@ Here is an example output:
...
+- `v`: version (to make data migrations easier in future)
+- `time`: local time stamp
+- `host`: host name
+- `cores`: number of cores on this node
+- `user` : username owning the process/command
+- `job`: job ID (0 if not applicable)
+- `cmd`: process/command
+- `cpu%`: CPU percentage (as they come out of `ps`)
+- `cpukib`: CPU memory used in KiB
+- `gpus`: GPU devices (more about it below)
```
lars-t-hansen, line 108: Documentation should specify the range of values; is it zero-based or 1-based?

README.md
```diff
@@ -86,22 +86,45 @@ Here is an example output:
...
-- CPU percentage (as they come out of `ps`)
-- memory used in KiB
+- `v`: version (to make data migrations easier in future)
+- `time`: local time stamp
+- `host`: host name
+- `cores`: number of cores on this node
+- `user` : username owning the process/command
+- `job`: job ID (0 if not applicable)
+- `cmd`: process/command
+- `cpu%`: CPU percentage (as they come out of `ps`)
```
lars-t-hansen, line 106: On a more general note I think we want more precision in this documentation.  Here's what I use elsewhere though I'd like to revisit some of these:

- version is on the format n.m.o
- time is ISO time without fractional seconds but with TZO
- host is FQDN
- cores is positive integer
- job id is positive integer; not convinced that zero should be legal here?
- user is not simply the user name, it can also be `_unknown_` and `_zombie_<something>`
- cpu% is percent of one core
- the card indices in gpus is 1-based (I'm not wedded to this; discuss)
- gpu% is sum across cards
- gpumem% is % of memory across all cards
- gpukib is sum across cards

bast:
Thanks for the feedback! I will adjust.

README.md
bast: I thought of allowing any string, not only numbers. But that would complicate "none" and "unknown". And for the numbers I thought of allowing whatever base is defined client-side.

lars-t-hansen: Re the number base, I sort of feel it's important to settle on one thing, because the analysis tool will not necessarily know what hardware the host has.  Although that is in itself an interesting issue, the current analysis tool does require information about the hardware configuration in some cases, and it would be possible to make better use of that, with care.

I suppose it's possible to leave it open here (ideally, explicitly so) and to account for that in the analysis tool, which does not currently assume a dense set anyway, and to tighten up here later if it becomes an issue.

README.md
lars-t-hansen: Reading the code some more I think gpu% and gpumem% and gpukib are sums across the cards that are used by the process whose information we're recording.  I'll look at this even more tomorrow.

bast: Good point about not allowing 0 job ID if it is not applicable. But then it should be positive integer or "none"? Or something like that?

lars-t-hansen: Having thought about this a little more, I guess "0" for "none" is better than a special token.  And "none" may be an unavoidable case in some scenarios.  On our systems without job queues, I run sonar with --batchless so that it synthesizes a nonzero job ID.  On systems with job queues, I guess "0" is something we use for processes that don't have job IDs in the queue.  I think I will be fine with that.  So consider my objection void.  (Unless you think there are batch queue systems that might use 0 for a valid job ID.)

README.md
lars-t-hansen: Yeah... if the fields are named they can be in any order, though, and my assumption with having named fields is that fields can be introduced and removed when needed.  And indeed a log file for a given host and day can have records with and without a particular field for that reason.  The ingestion code I have in the sonarlog library handles that properly (and by design).  I think we're going to be adding quite a number of fields over time and having defaults would be nice.

The alternative is running some migration pass every time we upgrade sonar.  This sounds unpleasant - pause all cronjobs on all nodes, upgrade all log files, upgrade sonar, restart cronjobs - and error prone.  But is this what you had in mind?

lars-t-hansen:
> Also if you could test the branch on the GPUs you have access to, that would help a lot. I haven't tested this outside of CPUs. But the tests are preserved.

Will do so once the discussion has settled down.

README.md
bast: I now also think that unordered named comma-separated fields is better than forcing it into a tabular format just for the sake of file suffix which in the example I should perhaps change away from `.csv`.

The version is there just in case somebody in future decides to migrate some data but I don't plan that on any regular basis. This is just to make it possible and us less worried about breaking format.

lars-t-hansen: The high bit is that we agree that the *syntax* is CSV with comma-separation and rules for quoting, so that we can use existing high-quality CSV parsers, even if the particular *semantic* detail of the tabular format is ignored.  Both Rust and Go have such parsers that appear to be flexible about the tabular format.

lars-t-hansen:
> > Also if you could test the branch on the GPUs you have access to, that would help a lot. I haven't tested this outside of CPUs. But the tests are preserved.
> 
> Will do so once the discussion has settled down.

The patch (as it stands) works fine with the GPUs.

README.md
lars-t-hansen: To close the loop on the gpu fields: they are samples (or close enough) and fine as it is.  The only problems we have are with cpu%, as discussed at length on bug #68.

bast:
This is still not done. I only resolved conflicts and rebased.

bast:
Implementing defaults is a good idea but postponed to #74 otherwise I will never complete this PR :-)

Remaining here:
- [ ] Change the output of `gpus` and then we do the rest as follow-up.
- [ ] Update README.md after recent changes.

bast:
I suggest we merge although this is far from perfect:
- Defaults are not properly implemented and documented but followed up in #74 
- `gpus` field is implemented as a HashSet and this will need to be modified for "unknown" and "none" (currently neither "unknown" nor "none" is implemented) followed up in #75
- Although the `gpus` code is now "wrong", tests don't catch it anyway. We will need to adapt the tests.
- I am pressed on deadlines and think that 80% OK code on `main` is better than PR never finished.

README.md
```diff
@@ -86,22 +86,46 @@ Here is an example output:
...
-2023-07-27T18:29:17+02:00,somehost,8,user,,brave,15.5,7085968
-2023-07-27T18:29:17+02:00,somehost,8,user,,.zoom,37.8,1722564
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=.vim-wrapped,cpu%=1.9,cpukib=7228,gpus=none,gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=node,cpu%=1.8,cpukib=79332,gpus=none,gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=slack,cpu%=0.7,cpukib=591720,gpus=none,gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=X,cpu%=1.5,cpukib=224416,gpus=none,gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=brave,cpu%=12.1,cpukib=3075300,gpus=none,gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=alacritty,cpu%=1.2,cpukib=286196,gpus=none,gpu%=0,gpumem%=0,gpukib=0
+v=0.7.0,time=2023-07-29T17:45:37+02:00,host=somehost,cores=12,user=someone,job=0,cmd=sonar,cpu%=9,cpukib=372,gpus=none,gpu%=0,gpumem%=0,gpukib=0
```
lars-t-hansen, line 95: These examples are incorrect (or at least incomplete) because there's now the cputime_sec field.  Also it would be very useful to show at least one example here with a nonempty set for gpus.

README.md
```diff
@@ -86,22 +86,46 @@ Here is an example output:
...
+- `cores`: number of cores on this node (positive integer)
+- `user` : username owning the process/command (it can also be "unknown" and "zombie")
+- `job`: job ID (positive integer; 0 if not applicable)
+- `cmd`: process/command
+- `cpu%`: CPU percentage (in percent of one core; as they come out of `ps`; this is not a sample but a running average)
+- `cpukib`: CPU memory used in KiB (this is a sample)
+- `gpus`: GPU devices (the card indices are 1-based; more about it below)
+- `gpu%`: GPU percentage (sim across cards; this is a sample)
+- `gpumem%`: GPU memory percentage (in percent of memory across all cards; this is a sample)
+- `gpukib`: GPU memory used in KiB (sum across cards; this is a sample)
```
lars-t-hansen, line 111: Add cputime_sec.

src/ps.rs
```diff
@@ -271,26 +272,33 @@ pub fn create_snapshot(
 
     let mut writer = Writer::from_writer(io::stdout());
 
+    const VERSION: &str = env!("CARGO_PKG_VERSION");
+
     for ((user, job_id, command), job_info) in processes_by_job_id {
+        // FIXME this does not print "none" or "unknown"
```
lars-t-hansen, line 278: Customarily, add a link to the filed issue with the FIXME.

src/ps.rs
```diff
@@ -271,26 +272,33 @@ pub fn create_snapshot(
...
+        let gpus_comma_separated: String = job_info
+            .gpu_cards
+            .iter()
+            .map(|&num| num.to_string())
+            .collect::<Vec<String>>()
+            .join(",");
+
```
lars-t-hansen, line 285: I think that it would be easy to add a line of code here so that if gpus_comma_separated is the empty string then it gets the value "none" instead, this is in keeping with the desired syntax and close enough to the eventual semantics that it is better than using the non-standard empty string in the output.

src/nvidia.rs
```diff
@@ -17,13 +16,13 @@ use crate::util::map;
...
 #[derive(PartialEq)]
 pub struct Process {
-    pub device: i32,         // -1 for "unknown", otherwise 0..num_devices-1
-    pub pid: usize,          // Process ID
-    pub user: String,        // User name, _zombie_PID for zombies
-    pub gpu_pct: f64,        // Percent of GPU /for this sample/, 0.0 for zombies
-    pub mem_pct: f64,        // Percent of memory /for this sample/, 0.0 for zombies
-    pub mem_size_kib: usize, // Memory use in KiB /for this sample/, _not_ zero for zombies
-    pub command: String,     // The command, _unknown_ for zombies, _noinfo_ if not known
+    pub device: Option<usize>, // -1 for "unknown", otherwise 0..num_devices-1
```
lars-t-hansen, line 19: The comment here is wrong now.

src/ps.rs
```diff
@@ -6,22 +6,22 @@ use crate::jobs;
 use crate::nvidia;
 use crate::process;
 use crate::util::{three_places, time_iso8601};
+use std::collections::{HashMap, HashSet};
```
lars-t-hansen, line 9: The location of this is a stylistic issue and we should probably agree on how we like to do it (and as the module owner you get to decide :-) but I've been grouping `use` directives in blocks in the following order: `extern crate X`; `use crate X`; `use X::Y`, with directives for tests following that in the same order.  I'm not enough of a rust programmer to know what is "typical" or "standard" in the community; happy for feedback.

lars-t-hansen (approved):
LGTM modulo comments that ought to be fixed and the tweak for writing "none" where it's easy to do so.

README.md
bast: Thanks! I will fix the example but will require you or a bit more time for a semi-realistic gpu line in there :-)

README.md
bast: Thanks! I will add.

src/ps.rs
bast: Good suggestion. Will add.

src/ps.rs
bast: Good workaround. I will add.

src/ps.rs
bast: I think this order makes sense. I will follow it.

PR #70, opened 2023-08-08
Add cputimes.  Fixes #68

### LINKED ISSUE #68: Print raw cpu time
Currently sonar logs the `pcpu` field from the `ps` output, but this is a pretty tricky number.  It is not a sample: it is the ratio of the process's consumed CPU time by the elapsed time since the process started, if I read the manual page correctly.  As time moves forward, it will take ever greater changes in process behavior to move this number at all.  (If a process sits around doing nothing for 24 hours and then runs 100% for an hour, this number will move from 0% to 2.5%.)  I think that what we instead want to log is the `cputimes` field, which is the consumed CPU time up until that point.  Given two consecutive log records we can then say something about how busy the process was during the last sampling interval, and it will be meaningful to look at averages, trends, and so on.  Some of these are core use cases for [NAICNO/Jobanalyzer](https://github.com/NAICNO/Jobanalyzer).

(Given enough precision in the output it may be possible to compute the desired value from the pcpu value, but ps prints only one digit after the decimal point and this is unlikely to be very good in practice.)

Once we have support for named fields in the sonar output (#41) we can easily add this field, and it shouldn't take any extra effort during sampling to generate the data.

Also see https://github.com/NAICNO/Jobanalyzer/issues/27.

lars-t-hansen:
Extract the `cputimes` field from ps and add it onto the end of the output, after the GPU fields.  Clients that can parse optional fields (such as the gpu fields) should not be disturbed by this.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -34,11 +34,17 @@ pub fn get_process_information(jobs: &mut dyn jobs::JobManager) -> Result<Vec<Pr
 
 const TIMEOUT_SECONDS: u64 = 2; // for `ps`
 
+// `--cumulative` and `bsdtime` are to make sure that the cpu time accounted to exited child
+// processes (the cutime and cstime fields of /proc/pid/status) is used and printed.  Note
+// `cputimes` is unaffected by `--cumulative`.
+//
+// The format of `bsdtime` is `m...m:ss` in minutes and seconds.
+
 const PS_COMMAND_FILTERED: &str =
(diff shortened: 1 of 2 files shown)


lars-t-hansen:
Converted to draft while I ponder some subtleties, see #68.

bast:
I just read up on the issue - it is really tricky.

bast:
Thank you!

PR #71, opened 2023-08-08
Better error reporting.  Fixes #52

### LINKED ISSUE #52: Log fatal errors
See #51.  We should detect unrecoverable error situations that prevent monitoring from working, and we should arguably log them to some standard syslog or other medium where they will be seen.

lars-t-hansen:
Basically:

- propagate errors from lower layers (haven't done anything about the slurm component yet) as meaningful strings
- carry the failed command and its output (if any) with the CmdError
- use the standard logger to log these errors

This change pulls in a fair amount of code, I think chiefly through the env_logger library used to initialize logging.

For my purposes, this logging (to stderr) is roughly what I want, since cron captures the output and emails it to me.  But it may get further tweaking down the line.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -407,6 +407,7 @@ dependencies = [
  "clap",
  "csv",
  "hostname",
+ "log",
  "num_cpus",
  "subprocess",
  "wait-timeout",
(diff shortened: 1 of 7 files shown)


bast:
Thanks for your work on this! At some point we will need to go through all the `unwrap()` calls but this is for later.

PR #77, opened 2023-08-10
Fixes #76

### LINKED ISSUE #76: pid is a unique key for results from the subprocesses
For ps, nvidia-smi, and rocm-smi (and in general for all future uses like these) the pid will be a unique key (for a given pid there is only ever a single user and a single command, at least per invocation of ps etc).  In ps.rs we still use the tuple (user, pid, command) as the key for various processing steps.  This is not necessary and obscures what happens.  It's probably just an effect of an older design that has yet to be cleaned up completely.

lars-t-hansen:
Fairly obvious rewrite.  Took the opportunity to remove some tuples, too, for additional clarity.

### DIFF
--- a/src/ps.rs
+++ b/src/ps.rs
@@ -62,9 +62,32 @@ fn add_job_info(
         });
 }
 
-fn extract_ps_processes(
-    processes: &[process::Process],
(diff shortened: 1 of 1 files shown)


bast:
Much nicer code. Thanks!

Issue #73, opened 2023-08-09
Job filtering should be post-accumulation, and must be more subtle


lars-t-hansen:
At the moment, sonar filters processes and jobs early based on the raw ps output: if the cpu percentage of the process itself is below a threshold or its current memory usage is below a threshold, then the record is discarded.  This is iffy for several reasons:

* multiple processes can be accumulated into a job and in total they may create a load that is above the cutoffs, yet none of those processes will be considered
* for `--batchless` in particular, but I suspect even in general, the root process of a job may not be doing much but it may well be that it should be present to represent the job as a whole
* the cpu percentage is, as we've seen in #68, not a high-quality number (and in particular it is not a sample)
* filtering, even when based on good samples such as for mem_percentage, creates a log that has holes, which can be surprising to log consumers, jobs will appear to have stopped and then restarted, and there can be confusion about whether a job ID seen after such a hole is the "same" job or a "new" job

I suspect that we should do multiple things here:

* filtering should be applied to the accumulated job records, if at all
* filtering might be disabled altogether for `--batchless`
* filtering by cpu% probably needs to be removed, but cputime is not much of a substitute since that is a cumulative value and only the consumer of the log can reconstruct a sample-ish value for it

It's possible this is a good time to discuss why we're doing filtering and what we're really interested in getting rid of.  Looking at my own logs, I'd be more interested in getting rid of jobs by root and zabbix than user processes, for example.

bast:
Thanks for good points!

My motivation for filtering was:
- smaller log files to not fill the disk (maybe I was too careful)
- smaller log files to have faster analysis later to not read through many zeros (also this is perhaps premature optimization)

lars-t-hansen:
Oh, and we filter twice and I forgot the other one: for non-`--batchless` operation we filter when we run ps.  This filtering is coarse and probably error prone, it's meant to match pcpu and pmem, and that again gets us into trouble with the meaning of the pcpu value, and it embeds an "expected" number of spaces, which is probably asking for trouble too.  Again I think it would probably be best to apply this filtering later.

lars-t-hansen:
Establishing an evolving task list:

- [ ] remove two instances of ad-hoc filtering from ps.rs, see above
- [ ] implement the option of filtering during or immediately before we write
- [ ] establish appropriate filters, eg, it may be useful to cutoff for little cpu or memory use, and it may be useful to filter by user name
[...]

PR #79, opened 2023-08-11
Revamp ps.  Fixes #78

### LINKED ISSUE #78: job_id is possibly a unique key for job information, but "it's complicated"
Follow-on from #76.

Currently the key for the "job" table in ps.rs is (user, job_id, command).  It's not completely clear why the user name or command are desirable here.

On the one hand, the job id should be a unique key for the job.

On the other hand: for a given job there could be many processes each with their own command, and with the command being part of the key each will have its own line in the sonar log.  This may be what we want (discuss), since it gives postprocessing a chance to see this.  (The alternative would be for the sonar log to be more complex and contain multiple commands.)  More far-fetched, each of those processes could have its own effective UID, so there could be multiple users, too.

On the third hand: it's a little strange that sonar would roll up multiple `python` processes into one record but distinguish this from, say, a `python3.9` process belonging to the same job.  In this case, it would be more natural for sonar to produce a record for each process, all of them tagged with the same job number.

This may become a discussion about what sonar should do, and what the postprocessor should do.

lars-t-hansen:
As discussed on #78, the use of user name and command name to distinguish process records and the attempted merging of process records that have the same user name and command name are vestiges of an older design.  This patch removes all of that and does the following:

- the pid is the unique key for a record internally
- information from multiple sources for the same pid are merged into the unique record for the pid
- the record is tagged with its job id but the job id is not a distinguishing characteristic of anything, internally
- as a result, there may be multiple records for the same job in the output, but this was always the case; now it is documented
- the intermediate "rolling up" of ps and gpu records with the same user, pid, and command is not necessary and has been removed
- it is documented that job id 0 in the output is special

So far as I can see, this does not materially change the volume of output data, which is determined mostly by filtering anyway.  I've kept the existing filtering here but moved it down to the output phase, so there may be very slight changes in the output produced, but this is not expected to be material, see #73 for longer discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,10 @@
 # sonar
 
 Tool to profile usage of HPC resources by regularly probing processes using
-`ps`.
+`ps` and other tools.
 
-All it really does is to run
-`ps -e --no-header -o pid,user:22,pcpu,pmem,size,comm`
-under the hood, and then filters and groups
-the output and prints it to stdout, comma-separated.
+All it really does is to run `ps` and other diagnostic programs under the hood,
+and then filters and groups the output and prints it to stdout, comma-separated.
 
 ![image of a fish swarm](img/sonar-small.png)
 
@@ -100,7 +98,7 @@ The columns are:
 - `host`: host name (FQDN)
 - `cores`: number of cores on this node (positive integer)
 - `user` : username owning the process/command (it can also be "unknown" and "zombie")
-- `job`: job ID (positive integer; 0 if not applicable)
+- `job`: job ID (positive integer; 0 if not applicable, see below)
 - `cmd`: process/command
 - `cpu%`: CPU percentage (in percent of one core; as they come out of `ps`; this is not a sample but a running average)
 - `cpukib`: CPU memory used in KiB (this is a sample)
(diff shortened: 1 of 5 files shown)


bast:
I just realized that for an MPI job with many tasks each task will have its own pid so that will multiply the output, right? Or is this filtered later?

bast:
EDIT: s/filtered/condensed/

lars-t-hansen:
That's right, and that's the main risk that I see here.  If a system uses a process-per-core model, and there are many cores on a node, there may be more output than before since we'll emit one line per process on the node, and previously we might not have.

If this is undesirable or an actual problem then we could consider rolling up multiple processes with the same job ID and command into a single output line.  We would do this just before the output stage.  We'd probably want to tag the record somehow to indicate that that has happened (eg, we could emit the number of records thus rolled up).  There would need to be a way to disable this because one of the use cases for Jobanalyzer is to look at balanced utilization, and that information would be lost when we roll up like this.  (That's basically the problem that sent me down the rabbit hole of cleaning up `ps` in the first place.)

However, note that if the jobs running on the node are *not* all part of the same MPI job but are various smaller jobs that happen to be scheduled on the node at the same time, then we'll get the full output volume anyway.

bast:
Good. Rolling up by job ID is a good plan. I just got worried that on machines like Betzy we will multiply output now by factor 100-1000.

lars-t-hansen:
OK, I'll add a tweak to do that.  I think I may prefer to let "no rolling up" be the default, as rolling up is an optimization, and it can be triggered by a command line switch in the same way that different filters can be.

bast:
Good idea to make this a non-default option.

lars-t-hansen:
An unexpected but sort of cool effect of `--rollup` is that many more records are printed on my local system.  This is because filtering is applied after rolling up.  So the 29 Firefox processes, which are mostly all below the cpu limit for printing, together get rolled into one record that is above the limit.  That record belongs to me, of course; all the other new records belong to root.  But this is fine - we're going to rethink filtering anyway.

lars-t-hansen:
@bast, PTAL when you have a spare moment (no real urgency).

bast:
Thanks a lot!

PR #80, opened 2023-08-11
Even better filtering.  Fix #73

### LINKED ISSUE #73: Job filtering should be post-accumulation, and must be more subtle

lars-t-hansen:
Major points:

- The filtering that we used to run directly after the ps subprocess has been removed
- The previous patch already moved filtering to the printing code in ps.rs, here we just expand on that
- We add a number of command line options to specify how jobs are filtered (more are possible but I figure we'll follow up with individual patches, driven by need)
- Existing command line options are cleaned up to create a coherent set and the filtering logic is itself tidied up
- We extract uid along with the user to more easily filter out system processes, this is fairly disruptive but mechanical
- Some minor drive-by cleanup
-

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 target/
 *~
+*.rej
+*.orig
--- a/src/amd.rs
+++ b/src/amd.rs
@@ -15,9 +15,9 @@
 /// device by the number of processes on the device.  This is approximate.
 use crate::command::{self, CmdError};
 use crate::nvidia;
+use crate::ps::UserTable;
 
 use std::cmp::Ordering;
-use std::collections::HashMap;
 
 #[cfg(test)]
 use crate::util::map;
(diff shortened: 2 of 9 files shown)


lars-t-hansen:
Oh, and the default is now "no filtering" and "no rollup".  I think this is roughly right but happy to discuss alternatives, in particular, it could be useful for there to be meta-switches to set defaults for different use cases.

bast:
Thanks! The new CLI and the new defaults sounds great. Thanks for the cleanup!

PR #81, opened 2023-08-11
Handle unknown gpus.  Fix #75

### LINKED ISSUE #75: For `gpus` implement "unknown"
The data is currently implemented as a HashSet and the output will be `gpus=` for "none" and "unknown" is not implemented.

lars-t-hansen:
Straightforward fix for #75, want to test it a little better first.  One can discuss whether the GPU sets aren't even more complicated, containing both some known and some unknown devices, but I'm not sure what we're going to get out of that.  Getting to the point where "there are some unknown devices here" is enough of an error indicator that it should trigger somebody to go and look closer.

### DIFF
--- a/src/ps.rs
+++ b/src/ps.rs
@@ -14,14 +14,40 @@ use csv::{Writer, WriterBuilder};
 use std::collections::{HashMap, HashSet};
 use std::io;
 
-type GpuSet = HashSet<usize>;
+// The GpuSet has three states:
+//
+//  - there are some gpus in the set, this is Some({a,b,..})
+//  - the set is known to be nonempty but have (some) unknown members, this is None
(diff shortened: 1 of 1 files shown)


lars-t-hansen:
This is looking good on one of our systems with a zombie job holding onto GPU resources, the GPU set is printed as "unknown" as desired.

lars-t-hansen:
```
v=0.7.0,time=2023-08-14T08:51:43+02:00,host=ml1.hpc.uio.no,cores=56,user=_zombie_3843643,job=0,cmd=_unknown_,cpu%=0,cpukib=0,gpus=unknown,gpu%=0,gpumem%=0,gpukib=968704,cputime_sec=0
```

bast:
Thanks!

PR #85, opened 2023-08-15
Exclude commands

### LINKED ISSUE #84: Add an option to filter by command name
Looking at the logs with the new code, they look pretty good but one thing that stands out is that, due to the new system with the cpu time for a terminated process being accounted to its parent, it may be useful for the sake of reducing log volume to allow filtering by command name.  The purpose would be to remove sshd / bash / zsh / systemd / tmux processes from the listing, they are not likely to be interesting though they pass other filters for the reason given.

Currently I run some tests with
```
--exclude-system-jobs --min-cpu-time=60 --batchless --rollup
```
and "shell" programs account for 2/3 of all output lines on ML8.  This is probably not representative for a batch system but it still seems like it'd be useful to do the filtering.

lars-t-hansen:
Fixes #84.  An easy change, but making it draft since it depends on #83, which is currently draft.


lars-t-hansen:
I broke my workflow again, will reopen in a clean repo.

Issue #74, opened 2023-08-10
Define, implement, and document defaults for values where it makes sense


bast:
> On that topic, a useful optimization for the log writer is the use of default values that are agreed between the logger and the ingestor. The default value for gpus would be none and sensibly the field does not appear in the output at all. For several other fields, a sensible default is zero and if the value is zero we could (optionally of course) omit the field. The ingestor will have to deal with missing fields and obsoleted fields and newly-introduced fields anyway, so this adds very little complexity.

Following up from #66.

bast:
Ideally we will encode these into the types.

lars-t-hansen:
Looking at the current output the sensible defaults might be (`rolledup` already has a default):

- cpu% => 0
- job => 0
- gpus => none
- gpu% => 0
- gpumem% => 0
- gpukib => 0
- cputime_sec => 0

With sensible command line switches the defaults really only matter for the gpu data and cpu%, since only longer-running true jobs and login shells will be printed.  For example, I run with `--batchless --exclude-system-jobs --exclude-users=larstha --min-cpu-time=10` and I get only those types of processes, none of which have job=0.  The `--min-cpu-time=10` switch ensures the cputime_sec default is unused.  Frequently, though, cpu%=0 because data are printed for very long lived login shells.

In practice the defaults will cut the output volume by 20-25%.  That may be worth it, or maybe not - discuss.  If we're concerned about output volume we could do vastly more.

PR #90, opened 2023-08-16
Print the pid field when it is meaningful

### LINKED ISSUE #31: Also record `pid` in the output of `sonar ps`

lars-t-hansen:
Fixes #31.


bast:
This is good and I will merge but just writing my thoughts for "later"
- there could be a real PID 0 also
- I am wondering whether we should print it at all if it is rolled up
- Later we can introduce types that map to this better like Option or even a custom type

But for the moment this is better than before, just thinking out loud.

lars-t-hansen:
>     * there could be a real PID 0 also

Oops.  The web seems to believe that on linux, pid 0 is sometimes the synthetic pid for the kernel, but that there is no /proc/0 and that 0 is also used for some other things.  Further investigations required I guess.  I will file a follow-up bug.

>     * I am wondering whether we should print it at all if it is rolled up

We should not, but that is part of #74.

>     * Later we can introduce types that map to this better like Option or even a custom type

I think that's also #74.

bast:
Perfect - thanks for clarification! I am reading things a bit out of order and I am sure there is a very good plan behind everything.

PR #98, opened 2023-08-21
Fix #97

### LINKED ISSUE #97: Sonar is slow without --batchless because of how we get slurm job IDs
Running a sonar release build just now on a lightly-loaded ML node (ml7, a beefy AMD system), it 0.27s real time with `--batchless` and in 2.5s real time without `--batchless`.  The difference is even more stark on my development system (a slightly older Xeon tower): 0.03s vs 1.63s.

I run with `--exclude-users=root --exclude-system-jobs --rollup` to keep the amount of output to a minimum, so that we can know it's not output generation that's the main problem.

Running `perf` on this it is clear that the problem is in `get_slurm_job_id`: *every* profiling hit in the first 40-line page of output is in the pipeline that that function runs to get the job ID.  We can probably do much better here (and we'll need to).

lars-t-hansen:
Avoid the subprocess for getting the slurm ID.  This massively speeds up the process of probing for and finding these IDs.  Somewhat tested locally on Fox, looks OK to me.

### DIFF
--- a/src/slurm.rs
+++ b/src/slurm.rs
@@ -1,8 +1,10 @@
 // Abstraction of jobs::JobManager for SLURM.
 
-use crate::command;
 use crate::jobs;
 use crate::process;
+
+use std::fs::File;
+use std::io::{BufRead,BufReader};
 
 pub struct SlurmJobManager {}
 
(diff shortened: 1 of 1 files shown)


bast:
Thanks!

PR #99, opened 2023-09-07
Fix #86

### LINKED ISSUE #86: Read data directly from /proc
Right now we're running `ps` but there are good reasons not to do so:

- needs to run a subprocess which needs to parse its command line, figure out what to do, obtain data, and generate textual data, which we then have to read and re-parse
- we need a lot of data from the subprocess to generate a process tree, which makes the previous problem quite a lot worse

By reading data from /proc/PID/status (mostly) we would streamline this process and avoid unnecessary work.

lars-t-hansen:
Read process data from /proc instead of running ps.  This change duplicates the action of ps without adding other possible optimizations, but even so seems to reduce the running time of sonar by about 50% (casual end-to-end testing with output going to /dev/null).  The semantics of ps are followed closely, the main discrepancy being roundoff errors for the pcpu and pmem fields, see comments in code.


bast:
This is amazing work! Thank you!

PR #104, opened 2023-09-22
more robust date query; closes #103

### LINKED ISSUE #103: Running `date` three times to compute the output file name is risky
The shell script [here](https://github.com/NordicHPC/sonar#how-we-run-sonar-on-a-cluster) runs date three times to generate a file name.  But if the first invocation is on yyyy-12-31 and the second invocation is on yyyz-01-01 this will compute the wrong name.  Ditto if the day changes between the second and third invocation, yielding yyyy-mm-31 and yyyy-mn-01.  Even though this is unlikely, date should ideally be run just once.

bast:
One `date` invocation instead of three.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,9 +290,7 @@ set -euf -o pipefail
 
 sonar_directory=/cluster/shared/sonar/data
 
-year=$(date +'%Y')
-month=$(date +'%m')
-day=$(date +'%d')
+read -r year month day <<< $(date +"%Y %m %d")
 
 output_directory=${sonar_directory}/${year}/${month}/${day}
 


README.md
```diff
@@ -290,9 +290,7 @@ set -euf -o pipefail
 
 sonar_directory=/cluster/shared/sonar/data
 
-year=$(date +'%Y')
-month=$(date +'%m')
-day=$(date +'%d')
+read -r year month day <<< $(date +"%Y %m %d")
```
lars-t-hansen, line 294: Alternatively
```
path=$(date '+%Y/%m/%d')
output_directory=${sonar_directory}/${path}
```

bast: Good idea. I will adjust.

lars-t-hansen (approved):

PR #108, opened 2023-10-02
Procfs tweak: ignore benign and expected errors

lars-t-hansen:
Since we first enumerate the /proc directory and only later walk the list of enumerated pids, we should expect that some of the pids may have disappeared in the meantime.  Right now this causes sonar (really procfs) to bail out, but it should not.  Convert some error returns into `continue`.


bast:
Thank you!

PR #110, opened 2023-10-04
Fix #72: Surface GPU failures in the log records

### LINKED ISSUE #72: Log GPU status
A reality of the GPUs is that they fail intermittently, either together or separately.  The GPU modules of sonar can (to a point) detect this and can at that point perform a slightly deeper analysis, and can add that analysis to the sonar log, this will likely be helpful in running a continuous health check.  The complexity for sonar as a whole is minimal: just a field devoted to GPU health, which can be absent by default and default to "all's well".

Just now I have a card failing on our ML2 node, and as a result the GPU data are not collected because nvidia-smi in its default mode reports that one of the cards could not be reached and bails out.  But digging slightly deeper we see it's just the one card
```
$ nvidia-smi -L
Unable to determine the device handle for gpu 0000:18:00.0: Unknown Error
GPU 1: NVIDIA GeForce RTX 2080 Ti (UUID: GPU-f0f3ae96-fe93-f5c8-4295-bf657e5a25da)
GPU 2: NVIDIA GeForce RTX 2080 Ti (UUID: GPU-99c4f278-11da-052d-ac5c-3490b58ade04)
GPU 3: NVIDIA GeForce RTX 2080 Ti (UUID: GPU-3c748ed6-c6b6-6ef0-ad7d-fe109cff16bc)
```
Packaging this information in the log somehow ("gpufail=<comma-separated list of failing gpus>") is probably useful.  This is not high priority.

lars-t-hansen:


bast:
Thanks!

PR #112, opened 2023-10-04
Fix #107: Always print at least one record

### LINKED ISSUE #107: Always write something
At the moment, `sonar ps` will not write anything if there are no jobs running worth reporting about.  However, for https://github.com/NAICNO/Jobanalyzer/issues/104 I need `sonar ps` to reliably write a record every time it is invoked, because the presence of data with the timestamp in the timeline is used as a signal that the system was up at that time.

Two obvious ways to fix this:
- A: `sonar ps` always writes at least something, even if only a synthetic record
- B: `sonar ps` invoked with a flag (say, --always-write) always writes at least something, ditto

Then, the secondary question:
- X: write a synthetic record always
- Y: write a synthetic record only if there is nothing else to write

Possibly there are natural affinities here (Y with A, X with B) but I'm uncommitted about that.  I'm most happy with Y.

Clearly there are some other options:
- Don't place this burden on sonar but use some other mechanism for uptime monitoring
- Make sonar do this, but log the data some other place than in the normal job data stream

Frankly though I think it fits in pretty well with whatever else sonar is doing, and since the sonar data require more or less sophisticated disentangling already anyway, and in the common case already contain the data we want for uptime monitoring, there doesn't seem to be any need to add more output files.
 
@bast, any opinions about this?

lars-t-hansen:
Draft, pending resolution to the discussion in #107, but ready to go if this is how we decide to do it.

This prints a synthetic heartbeat record from `sonar ps` if nothing else has been printed, and does not need a command line switch to do so - it always does it.  Easy to add a switch if we need it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -83,6 +83,7 @@ fn main() {
         } => {
             let opts = ps::PsOptions {
                 rollup: *rollup,
+                always_print_something: true,
                 min_cpu_percent: *min_cpu_percent,
                 min_mem_percent: *min_mem_percent,
                 min_cpu_time: *min_cpu_time,
--- a/src/ps.rs
+++ b/src/ps.rs
@@ -154,6 +154,7 @@ where
 
 pub struct PsOptions<'a> {
     pub rollup: bool,
+    pub always_print_something: bool,
     pub min_cpu_percent: Option<f64>,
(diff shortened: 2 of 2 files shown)


bast:
Thanks again so much for this improvement!

Issue #113, opened 2023-10-11
procfs can create data with infinite values


lars-t-hansen:
I've observed some unfortunate data in the sonar output (linebreaks inserted by me): 
```
v=0.7.0,time=2023-10-11T13:30:02+02:00,host=ml6.hpc.uio.no,cores=64,user=erlek,job=1685425,pid=0,cmd=python3,
cpu%=inf,cpukib=60728400,gpus=5,gpu%=48,gpumem%=26,gpukib=3721216,cputime_sec=508,rolledup=4
```
Observe the `inf` value for `cpu%` on line 2.  There were a number of records with this problem on the host in question, on the day in question (all for the same user).

This is probably caused by repeated roundoffs in procfs before that value is computed, but it could be something else.  Anyway it shouldn't occur.  The ps command computes this somewhat differently (output.c line 539), avoiding all this rounding, and in addition uses integer arithmetic, and in addition performs some cleanup during formatting.

See https://github.com/NAICNO/Jobanalyzer/issues/139 for more about the investigation.

lars-t-hansen:
Also see [this comment on #111](https://github.com/NordicHPC/sonar/issues/111#issuecomment-1779488532).

lars-t-hansen:
It's pretty tricky to test this because the architecture of the procfs module does not lend itself to testing - it goes directly to the file system for its data.  The trick is to pass an `fs` provider to procfs, and that can then provide test data in a test setting.  We only need fs::read_to_string and fs::read_dir, the latter is a little harder b/c we need to produce complex fs-owned data types, but we can instead provide a suitable abstraction for the few parts of the data we need.  Anyway this is more surgery than I really want to do right now.  The test case mentioned in #111 passes with my current patch, and the code is sensible, and diffing the output against ps output produces sensible differences (ie very few), so that's probably good enough.

lars-t-hansen:
I'll push a test for #111, but I think a test case for the present job eludes me - the code is "correct by construction", and I'm adding some more comments on the patch for #94 to make that even clearer.

PR #116, opened 2023-10-26
Fix #111 - skip dead jobs, add defunct tag

### LINKED ISSUE #111: procfs must be more clever about jobs in the X and Z states
Fun error message:
```
INFO: procfs failed: Could not parse sess from /proc/2839925/stat: 2839925 (python3) X 0 -1 -1 0 -1 4243524 0 0 0 0 0 0 0 0 20 0 0 0 10643829 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 17 3 0 0 0 0 0 0 0 0 0 0 0 0 0
```
sess is the sixth field, here -1.  In the manpage the format is %d, but the procfs code uses a usize parser.  The reason it's -1 is the X in the third field: the process is dead, we just caught it at the wrong time.

This particular problem is benign and we'd probably reject the record anyway, it has no useful information.

It's possible that the right thing here is:

- be more careful about using parsers that correspond with what the manpage requires
- handle out-of-range values in some sensible way
- maybe look to the state field to make sure the job is in a sensible state before deciding what to do next

lars-t-hansen:
This has a slight conflict with the patch for #113, but easy to resolve.

### DIFF
--- a/src/procfs.rs
+++ b/src/procfs.rs
@@ -108,7 +108,7 @@ pub fn get_process_information() -> Result<Vec<process::Process>, String> {
         let realtime;
         let ppid;
         let sess;
-        let comm;
+        let mut comm;
         let utime;
         let stime;
         if let Ok(line) = fs::read_to_string(path::Path::new(&format!("/proc/{pid}/stat"))) {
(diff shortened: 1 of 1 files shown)


bast:
Thank you!

Issue #94, opened 2023-08-17
Add `ps` test cases


lars-t-hansen:
The `ps` command is now almost perfectly parameterized: all it takes is for us to pass the output stream as a parameter instead of always going to stdout.  With that in mind, it should be easy to construct synthetic test cases for the command so that bugs like #93 can be avoided in the future.

lars-t-hansen:
And on that note, we really need to be able to mock the input too, see https://github.com/NordicHPC/sonar/issues/113#issuecomment-1780707052 for ideas.

PR #117, opened 2023-10-26
For #94 - virtualize procfs, add tests

lars-t-hansen:
This is the first step.  procfs is virtualized completely so that it's possible to inject data into it (easily, too).


bast:
Thank you!

PR #118, opened 2023-10-26
Cosmetics

bast:
I ran `cargo fmt` and applied almost all suggestions from `cargo clippy` except one where I did not understand the code and the suggestion but that's for later.

### DIFF
--- a/src/ps.rs
+++ b/src/ps.rs
@@ -300,7 +300,7 @@ pub fn create_snapshot(jobs: &mut dyn jobs::JobManager, opts: &PsOptions) {
     // as not all records from this sonar run are filtered out by the front end.
 
     if gpu_status != GpuStatus::Ok {
-        for (_, proc_info) in &mut proc_by_pid {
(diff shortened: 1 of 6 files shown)


src/ps.rs
```diff
@@ -455,35 +460,42 @@ fn print_record<W: io::Write>(writer: &mut Writer<W>, params: &PrintParameters,
...
-            included = false;
-        }
+    if !params.opts.exclude_users.is_empty()
+        && params
+            .opts
+            .exclude_users
+            .iter()
+            .any(|x| *x == proc_info.user)
+    {
+        included = false;
```
lars-t-hansen, line 470: This change and the one below probably make the code strictly harder to read...

lars-t-hansen (approved):
Generally an improvement, so +1 for doing this.  The most annoying thing to me right now is probably that `cargo fmt` insists on surrounding infix binary operators with spaces, and `go fmt` insists on removing them, so it's impossible to have a consistent visual style across the two languages.

PR #120, opened 2023-10-27
Fix #100 - obtain timestamp early

### LINKED ISSUE #100: Maintain monotonicity of timestamps even in the face of some failures
For the background, see https://github.com/NAICNO/Jobanalyzer/issues/63, where timestamps did not properly reflect the time of measurement, but the time of reporting, and where measurements that were ordered by timestamps were not ordered by true time.  The reason for this is ultimately a system problem (program was hung waiting for disk and many instances were queued up and ran out of order, generating out of order timestamps) but we can mitigate it.

It might be to our advantage to obtain the timestamp for the output records very early, so that if jobs are backed up for system reasons, an output record with a lower timestamp has older measurements than a record with a higher timestamp.  (The order of the records in the output file does not matter, at least not to sonalyze, but internal consistency matters.)  This probably means obtaining the timestamp in `main()` even before initializing the logger.  This might still not be early enough if the system is stuck waiting for disk, but it'll likely be an improvement.

lars-t-hansen:


bast:
Thank you for the change. I am working on resolving the conflict and will then merge.

PR #124, opened 2023-11-07
Fix #123 - implement a lockfile system

### LINKED ISSUE #123: Guard against multiple concurrent invocations
As a prerequisite for deploying Sonar on Fox, the admins are requesting that there be a guard against sonar processes hanging and using up system resources, and are suggesting we use a lockfile for that.  As this is a feature for a subset of installations and not all installations, I imagine a command line option to `sonar ps` along the lines of `--lockfile /path/to/lockfile/dir`.  It would be ideal for the lockfile dir to be on a local file system but it would be best for us not to require that, so the actual lockfile name would probably be something like `/path/to/lockfile/dir/sonar-lock.<hostname>` where the hostname would be the host's detected host name.  The semantics would roughly be that sonar exits immediately if that file exists when sonar starts up (I need to read up on best practices).  Open question: What to do if the directory does not exist.

On the one hand, sonar already has some guards against hanging - we time out on subprocesses and so on.  On the other hand, we have already experienced hanging and some unfortunate fallout from that, see bug #100.

To test that this works it'll be useful to have some sort of testing mode that makes sonar sleep for a bit, and maybe an exit code from sonar that exposes the early exit.

lars-t-hansen:
Add a `--lockdir` option and attempt to create a lockfile in that directory if the option is provided.  Exit early if the lockfile can't be created.  Error out under various conditions, notably if the directory does not exist.


bast:
Thanks a lot! I need to look through it during a free minute.

lars-t-hansen:
No rush.  The patch has been deployed on Fox for testing and I'll let you know if anything strange happens.

lars-t-hansen:
Probably want a couple more things:

- signal handling so that at least SIGHUP, SIGINT, and other common signals that can cause termination won't cause sonar to exit without cleaning up the lockfile first
- clear documentation that if the option is used then there needs to be a cron job that cleans up lockfiles on reboot, minimally, and how to do that (notably, what will the file name be like)

bast:
Sorry for the silence but now I am back and looking at this. Thinking ...

lars-t-hansen:
I'm coming back this this now.  So far there have been no problems with the current solution, which has been deployed on Fox for quite some time, but I'll be implementing the two features in my previous comment.  @bast, If you have other opinions about the solution this is a great time to air them :-)

lars-t-hansen:
Cleaning up lockfiles on reboot can be done a couple of ways:

- place the lockfile in /run or /var/run, this is sensible esp if the sonar user's homedir is /var/run/sonar, as it probably ought to be
- place the lockfile on a tmpfs
- write the boot time of the system to the lock file and if the lock file exists, check that the boot time corresponds to current boot time and if not delete the file, then retry the acquire (elaborate)
- add an `@reboot` cron job to delete the file

Storing it on /var/tmp as I do on the current Fox deployment without further mitigations is *not* advisable.

bast:
Thanks for implementing this and for patience!

PR #131, opened 2024-01-12
Fix #126 - report total memory and per-job rss

### LINKED ISSUE #126: More memory data
I'm seeing a lot of memory overallocation on some of our systems (500% memory occupancy reported by sonar) and I'd like to know more about it.  I don't think sonar is wrong but I need to know what kind of artifact it is.  We should consider measuring more memory data, specifically, we probably want to look at resident set size and paging activity per process, but possibly also system-wide - to be explored further.

lars-t-hansen:
This is a WIP, given that rss isn't quite the number we want for rolled-up processes at least.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -487,7 +487,7 @@ dependencies = [
 
 [[package]]
 name = "sonar"
-version = "0.7.0"
+version = "0.8.0"
 dependencies = [
  "chrono",
  "clap",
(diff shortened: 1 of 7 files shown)


bast:
Sorry for conflicts ...

lars-t-hansen:
Closing this in favor of another patch set.

PR #133, opened 2024-01-29
Fix #126 - report resident memory

### LINKED ISSUE #126: More memory data

lars-t-hansen:
Four patches in this queue but only the last one (Compute resident memory) should have any drama, the others are cleanup that became necessary/desirable.  Probably worth mentioning that the "Refactor memtotal" patch adds an output field for RAM installed (to complement the core count field) and increases the version number to 0.8.0.

For resident memory, I currently go for the RssAnon fields of /proc/pid/status (and the drs field of the ps output, which is somewhat closely related), and the output field is called "rssanon".  Should we choose to print more fields or other fields we can then name them precisely.  This is better than having a definition that shifts or a name that is easy to misinterpret.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -487,7 +487,7 @@ dependencies = [
 
 [[package]]
 name = "sonar"
-version = "0.7.0"
+version = "0.8.0"
 dependencies = [
  "chrono",
  "clap",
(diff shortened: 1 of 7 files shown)


bast:
Thank you very much!

PR #136, opened 2024-02-12
Fix #134 - remove ps fallback.

### LINKED ISSUE #134: Remove `ps` fallback?
The use of /proc for performance data seems to be working out well, and we could remove the fallback that runs `ps` to obtain the same data.  That is increasingly appealing as `ps` can't quite supply the data we can get out of /proc.

lars-t-hansen:
- remove the ability to run `ps`, the code that examines `/proc` seems robust enough
- update the documentation some more, both to reflect this and to make changes that should have been fixed for #126.


lars-t-hansen:
Grumble.

bast:
Thanks! I will just resolve conflicts and merge.

PR #137, opened 2024-02-12
Fix #132 - don't crash if writing fails

### LINKED ISSUE #132: sonar crashes when output is not all consumed
```
$ ~/p/sonar/target/release/sonar ps --exclude-system-jobs --rollup | head -n 10
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error(Io(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }))', src/ps.rs:485:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

lars-t-hansen:


bast:
Thanks! I will just resolve conflicts and merge.

bast:
There is this conflict:
```
fn print_record<W: io::Write>(
    writer: &mut Writer<W>,
    params: &PrintParameters,
    proc_info: &ProcInfo,
<<<<<<< HEAD
) -> Result<bool> {
    let gpus_comma_separated = if let Some(ref cards) = proc_info.gpu_cards {
        if cards.is_empty() {
            "none".to_string()
        } else {
            cards
                .iter()
                .map(|&num| num.to_string())
                .collect::<Vec<String>>()
                .join(",")
        }
    } else {
        "unknown".to_string()
    };
||||||| 78d9dba
) -> bool {
    let gpus_comma_separated = if let Some(ref cards) = proc_info.gpu_cards {
        if cards.is_empty() {
            "none".to_string()
        } else {
            cards
                .iter()
                .map(|&num| num.to_string())
                .collect::<Vec<String>>()
                .join(",")
        }
    } else {
        "unknown".to_string()
    };
=======
) -> bool {
    // Mandatory fields.
>>>>>>> main
```

I believe here we want to replace `bool` with `Result<bool>` but we also want to remove the `let gpus_...` call like it was done on `main`?

lars-t-hansen:
I'll tidy it up and post a new patch.

lars-t-hansen:
@bast PTAL, should be OK now.

bast:
Thanks!

PR #140, opened 2024-02-13
Fix #102 - codify versioning policy + cleanup README

### LINKED ISSUE #102: Bump the version number for other things than file format changes
As we're getting closer to deploying sonar in earnest - testing so far proves that it has real utility - it will be helpful for us to update the version number more frequently, so that we can more easily track the version that's being used.  For example, the recent upgrade to reading data from /proc warranted such a change (though we did not make it).  I guess we'll have to figure out some sensible policies for when to bump the version.

lars-t-hansen:


bast:
Thanks! It's really good to get this formalized.

PR #146, opened 2024-02-22
Fix #142 - implement sysinfo

### LINKED ISSUE #142: Add `sysinfo` verb
Right now I have a standalone `sysinfo` program (https://github.com/NAICNO/Jobanalyzer/blob/main/code/sysinfo/sysinfo.go) that queries the configuration of the node and prints the data on stdout (as json), from where it is forwarded to the analysis machine.  On the ML nodes it is now run on boot + once a day.  I would like to have fewer programs to upgrade on all the nodes (currently there are three + a script and I'd like to reduce that to one + a script) and I think folding `sysinfo` as a verb into sonar makes a lot of sense.  It would not be much code (source or binary), we have most of the bits already.

lars-t-hansen:
This seems to be feature complete and working, I just need to make a final pass before asking for review.

### DIFF
--- a/src/nvidia.rs
+++ b/src/nvidia.rs
@@ -46,7 +46,7 @@ pub fn get_nvidia_configuration() -> Option<Vec<gpu::Card>> {
                                 if let Some(m) = model_name {
                                     cards.push(gpu::Card {
                                         model: m,
-                                        mem_size_kib: n * 1024 * 1024,
+                                        mem_size_kib: n * 1024,
                                     });
                                     model_name = None;
                                 }
--- a/src/nvidia.rs
+++ b/src/nvidia.rs
@@ -7,6 +7,7 @@
 /// Crucially, the data are sampling data: they contain no (long) running averages, but are
 /// snapshots of the system at the time the sample is taken.
 use crate::command::{self, CmdError};
+use crate::gpu;
 use crate::ps::UserTable;
 use crate::util;
 use crate::TIMEOUT_SECONDS;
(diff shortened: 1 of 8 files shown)


lars-t-hansen:
Also some test cases would be good.

src/nvidia.rs
```diff
@@ -7,31 +7,76 @@
...
-    pub mem_size_kib: usize,   // Memory use in KiB /for this sample/, _not_ zero for zombies
-    pub command: String,       // The command, _unknown_ for zombies, _noinfo_ if not known
-}
+// `nvidia-smi -a` dumps a lot of information about all the cards in a semi-structured form,
+// each line a textual keyword/value pair.
+//
+// "Product Name" names the card.  Following the string "FB Memory Usage", "Total" has the
+// memory of the card.
+//
+// Parsing all the output lines in order yield the information about all the cards.
```
lars-t-hansen, line 24: "yields"

src/sysinfo.rs
```diff
@@ -0,0 +1,109 @@
...
+use serde::Serialize;
+
+use std::io::{self, Write};
+
+pub fn show_system(timestamp: &str) {
+    let fs = procfsapi::RealFS::new();
+    match do_show_system(&fs, timestamp) {
+        Ok(_) => {}
+        Err(e) => {
+            println!("FAILED: {e}");
```
lars-t-hansen, line 16: Should use normal log channel, not stdout!!

src/sysinfo.rs
```diff
@@ -0,0 +1,109 @@
...
+        // Merge equal cards
+        let mut i = 0;
+        let mut gpu_desc = "".to_string();
+        while i < cards.len() {
+            let first = i;
+            while i < cards.len() && cards[i] == cards[first] {
+                i += 1;
+            }
+            let memsize = if cards[first].mem_size_kib > 0 {
+                format!("{}", cards[first].mem_size_kib * 1024 / GIB as i64)
```
lars-t-hansen, line 74: .to_string() instead of format?

bast:
Thank you!

PR #150, opened 2024-02-26
Fix #149 - Don't print cores and memtotalkib

### LINKED ISSUE #149: Remove cores and memtotal from `sonar ps` output?
With `sonar sysinfo` available, the `cores` and `memtotal` fields of the ps output are redundant - the information can be had from the sysinfo output, should it be needed.  Sonalyze does not use these fields of the sonar output at all, as it relies on the more complete config information.  (See also #105.)  I was afraid jobgraph might use `cores` but it does not.  So it would seem that:

- these fields can be removed from the `sonar ps` output, which is easy now that fields are named and have default values
- the sonarlog logic that reads the fields can be removed

@bast, WDYT?

lars-t-hansen:


bast:
Is it still work in progress? Should I look at the conflicts and merge?

lars-t-hansen:
Good question!  I think I was planning to test `sonar sysinfo` more thoroughly first and build up a little infra around that.  Currently doing so.

Also, it's a little bit up to you what you need for jobgraph and any other clients you're working on.  If it's more convenient for jobgraph (say) to have all the info in one place then I don't need to remove all these fields.  Hence the "WDYT" on #149.

lars-t-hansen:
I think this is good to go, `sonar sysinfo` works in production here.  But you might still consider whether this change is something you want.

bast:
I will look ASAP (another teaching week ...).

Issue #152, opened 2024-04-02
Supply chain security / simplify dependencies


lars-t-hansen:
This has been on my mind for a while but the recent exploitation of xz makes it a good time to talk about it.  Sonar runs (or will run) on every node on almost every supercomputer in Norway.  It pulls in not too many dependencies but some of them (clap, serde, libc) are quite large and actively maintained and some (subprocess, libc) are expected to do dark things by their nature.

https://news.ycombinator.com/item?id=39902241
https://research.swtch.com/xz-timeline

serde is an interesting case in point because for a while (possibly still but I've not been able to substantiate that) it shipped a precompiled binary in order to speed up compilation time, https://github.com/serde-rs/serde/issues/2538.  With this type of setup, vetting what we pull in and how it works is very hard.

Sonar actually has very simple needs and most of its libraries have been pulled in not because we need the full functionality but only a small part of it.  It may be to our benefit to go over the list of dependencies, remove what we can, and perhaps audit and pin what we can't remove, and define some kind of upgrade protocol.  For example, we have simple needs for csv and json production and for command line parsing and could easily replace those libraries if we wanted to.

I'm uncertain about how worried to be about this.  Sonar *currently* runs with no special privileges and its only superpowers are that (a) it runs regularly and everywhere and (b) if an admin sees it running, she will think it is fine that it's running.  It is otherwise no worse than any user program, indeed user programs whose source and binary we do not control.  On the other hand, being everywhere and running all the time under the radar it is a very convenient stepping stone for a more sophisticated attack leveraging other mechanisms.  And it is not out of the realm of the possible that Sonar will need some special powers eventually, for some types of monitoring.  And if not Sonar, then maybe a program that Sonar runs - same problem.

bast:
Great point. Let's go together through list of dependencies and let's check what we use them for and how much work it would be to replace some of them.

lars-t-hansen:
List of dependencies with uses, and some preliminary commentary:

chrono = "0.4"
* `chrono::Local::now()` for local time
* looks like this delegates to std::SystemTime and our use could be replaced by that

clap = { version = "4.5", features = ["derive"] }
* argument parsing and argument error reporting, but our needs are very simple
* hand-writing this would be easy

csv = "1.3"
* csv writer to ensure proper quoting, but our needs are very simple
* hand-writing this would be easy

env_logger = "0.11"
* `env_logger::init()`
* see `log` below, at least one of the two is redundant

hostname = "0.3"
[...]

PR #153, opened 2024-04-03
For #152 - remove wait-timeout dependency

lars-t-hansen:
The wait-timeout library is no longer used by sonar, it's a dead dependency in Cargo.toml.


bast:
Thanks!

PR #154, opened 2024-04-03
For #152 - Remove serde_json dependency

lars-t-hansen:
This removes serde_json and the direct dependency on serde (though serde is still included indirectly via csv).

### DIFF
--- a/src/sysinfo.rs
+++ b/src/sysinfo.rs
@@ -5,14 +5,14 @@ use crate::gpu;
 use crate::nvidia;
 use crate::procfs;
 use crate::procfsapi;
+use crate::util;
 
-use serde::Serialize;
-
-use std::io::{self, Write};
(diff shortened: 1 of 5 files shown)


src/sysinfo.rs
```diff
@@ -90,22 +75,32 @@ fn do_show_system(fs: &dyn procfsapi::ProcfsAPI, timestamp: &str) -> Result<(),
...
+"#);
+
+    // Ignore I/O errors.
+
+    let _ = writer.write(s.as_bytes());
+    let _ = writer.flush();
+    Ok(())
 }
+
+// Currently the test for do_show_system() is whitebox, see ../tests.  The reason for this is partly
```
lars-t-hansen, line 104: Mm, "whitebox" should be "blackbox" here.

PR #162, opened 2024-04-04
Fix #152 - Bump version, documentation, tidy up

### LINKED ISSUE #152: Supply chain security / simplify dependencies

lars-t-hansen:
Evolving - to land after everything else has landed.


bast:
After I merge and resolve the other PRs, I will fix conflict here and also update the lock file and add it to this PR.

bast:
Conflict fixed and Cargo.lock updated.

bast:
Before merging, let's run `cargo clippy`. It indicates few potential problems not caught by `cargo check`.

lars-t-hansen:
Formatting might be in order, too.

lars-t-hansen:
I'll pull it over to a machine with GPUs to make sure things look OK there too.

lars-t-hansen:
I introduced a bug in command line parsing when I removed clap.  Previously for sonar one could write `--option=value` not just `--option value`.  The scripts use the former syntax and I think it's useful to allow that.  I'll create a fix.  To keep things simple, I'll attach the fix to this PR.

lars-t-hansen:
That seems to work fine.  I'm going to think about whether we can have some test cases to test command line parsing.

lars-t-hansen:
I guess I went a little crazy with the testing.  But it's time to review this probably.

Cargo.toml
```diff
@@ -7,4 +7,4 @@ edition = "2021"
 
 [dependencies]
 libc = "0.2"
-subprocess = "0.2"
+subprocess = { version = "= 0.2.9" }
```
bast, line 10: it would have been sufficient to do: `subprocess = "0.2.9"`

bast:
Are the tests run as part of the workflow/pipeline/actions?

Cargo.toml
lars-t-hansen: Are you sure?  In https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html we read about the version "0.1.12":

> Although it looks like a specific version of the time crate, it actually specifies a range of versions and allows [SemVer](https://semver.org/) compatible updates. An update is allowed if the new version number does not modify the left-most non-zero number in the major, minor, patch grouping. In this case, if we ran `cargo update time`, cargo should update us to version 0.1.13 if it is the latest 0.1.z release, but would not update us to 0.2.0.
...
> This compatibility convention is different from SemVer in the way it treats versions before 1.0.0. While SemVer says there is no compatibility before 1.0.0, Cargo considers 0.x.y to be compatible with 0.x.z, where y ≥ z and x > 0.

What I was trying to express here was that `cargo update` should not be allowed to update this crate, because we do not trust it, and since Cargo does Weird Things before version 1.0.0, my understanding was that the "=" is required in this case.  Had the module been at 1.0.0 or higher you would indeed be right.

lars-t-hansen:
> Are the tests run as part of the workflow/pipeline/actions?

They are not, for two reasons (that are mostly the same reason).  One, I don't know how to make that happen :-)  Two, one of the tests has a dependency on an external tool `jq` to syntax check JSON, and I don't know how to set that up.

The dependency on `jq` can probably be fixed, if it turns out to be a problem.  It was an expedient solution.  We could create simple, custom testing programs for syntax checking JSON and CSV, and use those here.  Indeed the Jobanalyzer repo already has one of those that I wrote before discovering `jq`.

It would be great if these tests could be run as part of the workflow, though.

bast:
OK I will try to add those to our workflow.

lars-t-hansen:
Rebased.

Cargo.toml
bast: Ah! Thanks for explanation! I visually missed the "=" inside the quotes.

lars-t-hansen:
> Working on adding the tests ...

That's great.  When you get to a point where it works, also note some tests landed with #169 that should be added to the test runner.

bast:
Rebasing ...

bast:
The test fails now but for good reasons.

bast:
Can you please verify 284d507? This restores one shell test but I want a second pair of eyes that I did not misunderstand something.

lars-t-hansen:
Yes, that looks right - --bartchless and --rollup are incompatible now, so splitting the test is the best fix.

bast:
Now something else fails. But it passes on my laptop. Looking into it ...

bast:
Now we are all green finally. Ready to go from my perspective.

lars-t-hansen:
I'll take a quick look.

lars-t-hansen:
LGTM, merge at will.  I don't think I had anything outstanding on this one, getting rid of subprocess would be nice but I think that's a separate Issue.

PR #164, opened 2024-04-10
Fix #88 - spawn subprograms directly, not via shell

### LINKED ISSUE #88: Use exec, not shell, to run subprograms
See #86 and #87 also; full fixes for those makes the following point moot.

Sonar uses the `shell` command to run subprograms.  This forks an intermediate shell with all its baggage that then runs the subprogram for us.  This is not really necessary (it adds overhead); we could be using fork+exec or similar functionality to invoke the subprograms directly.

lars-t-hansen:
Improves efficiency, makes accidents harder, and sets us up for getting rid of the subprocess library.

### DIFF
--- a/src/amd.rs
+++ b/src/amd.rs
@@ -40,7 +40,7 @@ use crate::util::map;
 // too small.  This is presumably all driver dependent.)
 
 pub fn get_amd_configuration() -> Option<Vec<gpu::Card>> {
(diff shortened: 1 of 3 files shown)


lars-t-hansen:
Passes tests on AMD and NVIDIA hardware too, so this looks good to go.

PR #165, opened 2024-04-10
Fix #89 - delineate batchless jobs properly

### LINKED ISSUE #89: Unlikely peak CPU usage for jobs with --batchless and subjobs
Here's a funny one.  The fifth field (26268) is peak CPU usage, corresponding to 262 cores going at 100%:
```
957985	joachipo  0d 0h35m  3284 	26268 	1    	1     	0    	0     	0       	0        	ml3   bash
```
This is on a system with 64 cores, so it's clearly wrong: the max peak is 6400.

The reason this happens is an artifact of how `--batchless` jobs are handled by sonar.  In this case, processes below a session leader are taken to be root processes for individual jobs (which could be trees of processes); typically the shell is a session leader.  So if somebody runs five (say) heavy python processes (or process trees) in sequence from the shell, each of those is a job, as desired.  Because of how we collect the `cputime_sec` field as the sum of the process's own CPU time and that of all terminated children, however, the CPU time for all those jobs is eventually accumulated in the shell.  It can thus happen that when a long-running job terminates the shell receives a tremendous increase in accumulated CPU time out of nowhere within the time window between two samples.  Since the analysis code computes cpu utilization as (delta cpu usage)/(delta time), the usage can become very high, leading to the artifact shown above.

This is not a big deal at this time, as it pertains only to the shell and its relation to jobs below it, and really we don't care about the shells for analysis, so we can just sweep these artifacts under the carpet.  But it really comes back to how our aggregating data collection leads us into tight corners.  We chose to use the bsdtime field of the ps output along with the `--cumulative` switch to collect self+children time, and this is right for individual jobs, but it breaks down in this weird cross-job situation.

There are a couple of possible fixes, none of them urgent (details need to be worked out in both cases):

- for processes that are session leaders, collect only self time, not self+children
- do not aggregate data the way we do but instead present it in a way that lets the consumer aggregate according to need, for example, by a sonar record containing a list of processes with the self cpu time

lars-t-hansen:
This is probably roughly right but needs more testing for sure.  In local testing with the TC on the bug it performs OK, but that's a simple case.


lars-t-hansen:
This is being replaced by other code.

PR #174, opened 2024-05-21
Fix #173 - format tzo correctly

### LINKED ISSUE #173: Regression: Timestamp format is wrong
A subtle bug in the rewrite of sonar: it formats the timestamp incorrectly.  The correct format is `yyyy-mm-ddThh:mm:ss+hh:mm` but strftime uses `...+hhmm`.  Previously we did not have a test case.  Now we do, but because it assumed that the format was correct it is actually wrong.

lars-t-hansen:
I have not updated the version number here but it's possible we should bump the bugfix version - your call.

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -1,6 +1,6 @@
 use std::ffi::CStr;
 
-// Get current time as an ISO time stamp: yyyy-mm-ddThh:mm:ss+hhmm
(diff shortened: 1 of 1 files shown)


bast:
It probably does not matter since not too many users yet but I will bump the version in this PR just so that we start a good process. I will do it and then merge. Thank you!

PR #179, opened 2024-06-20
Fix #178 - extract numa info

### LINKED ISSUE #178: Include NUMA information
From a discussion with Marcin about Sonar on Betzy:

Include individual core utilization (high/low) as mask in ps data: https://github.com/NAICNO/Jobanalyzer/issues/508
Include NUMA architecture info in sysinfo data: https://github.com/NAICNO/Jobanalyzer/issues/509

The utility of the data is to be able to look for wildly unbalanced jobs automatically.

lars-t-hansen:
Currently a WIP, decoding /proc/stat and outlining various representation strategies for the data to make them compact.  There are endless possibilities, but a base-45 variable-length delta-from-minimum encoding with separators represented as distinguished leading digits is probably about as compact as more standard encodings (eg uleb64+base64), and simpler than the latter.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -9,6 +9,7 @@
       "name": "Bockwoldt, Mathias"
     },
     {
+      "affiliation": "University of Oslo",
       "name": "Hansen, Lars T."
     },
     {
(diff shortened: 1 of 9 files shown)


lars-t-hansen:
I should probably add an integration test too now that there is a command line option.

bast:
Thank you!

PR #189, opened 2024-10-07
Fix #44 - check for presence of GPUs before running probes

### LINKED ISSUE #44: Add a pre-check whether host/node is GPU-aware before running any other GPU commands

lars-t-hansen:


lars-t-hansen:
The failing test here looks like it may be a problem with the test setup.  The test passes locally.  It might fail if $USER is not set in the environment, which is a credible possibility for the test bot.

lars-t-hansen:
OK, this is a little bit bad.  The failing test was tests/ps-syntax.sh.  It failed because the user name that was seen did not conform to the pattern in the test.  I don't know what the user name in question was because there was no logging of that.  I have added logging now, but of course now the test passed so I have no info.  Presumably the user name is chosen from some character set that is larger than [a-z0-9]+, I'm guessing uppercase, or underscore or other special character.  The next time this happens we should get more actionable info.

lars-t-hansen:
Bugfix for tests/ps-syntax.sh landed separately on #193.

lars-t-hansen:
Backported to release_0_12.

PR #190, opened 2024-10-07
Fix #101 - print version and repo URL for 'help'

### LINKED ISSUE #101: `sonar help` should print version and build date
I think it will be helpful for sonar to be able to self-identify.

lars-t-hansen:


lars-t-hansen:
Same failure as on #44.

PR #195, opened 2024-10-08
Fix #168 - get rid of all cases of unwrap()

### LINKED ISSUE #168: Get rid of `unwrap`
There are a number of `unwrap` calls in the code.  Most of these are probably just fine b/c we've already checked the condition that must hold for the result to be something, but we could/should probably replace them with `expect` with a suitable string to document that.

lars-t-hansen:
In test cases, use `expect` instead of `unwrap`, with a string that clearly marks it as a test case.

In most other cases, use `expect` if there was an existing check on the path (typically `is_some`) and otherwise rewrite the code using `match`.

In a couple of cases, the `expect` checks a complex precondition and now properly documents that.

There are two cases of `panic` that have been commented but left alone.

### DIFF
--- a/src/amd.rs
+++ b/src/amd.rs
@@ -175,7 +175,7 @@ PID 28154 is using 1 DRM device(s):
     let users = map! {
     28156 => ("bob", 1001usize)
     };
-    let zs = extract_amd_information(concise, pidgpu, &users).unwrap();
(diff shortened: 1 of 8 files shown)


bast:
Thank you! This is an important improvement.

PR #197, opened 2024-10-12
Fix #96 - add more information to errors in subprocess execution

### LINKED ISSUE #96: Improve error logging further
See #92 for some background: there are some error codes such as "Hung" (but really most of them) that can have multiple root causes, we should include these root causes in the error message for smoother diagnosis of problems, or we should have more error codes.  (The latter complicates the API though.)

lars-t-hansen:
Baby steps, but at least something.


bast:
Thank you!

PR #203, opened 2024-10-29
Fix #200 - Extract more GPU information for sysinfo

### LINKED ISSUE #200: Collect more sysinfo data
At least these:

- GPU power level
- GPU clock speed, other things
- GPU interconnect
- GPU model

It may be that some of these (power level) change often enough that sysinfo is not the right venue for it but it's a start.

lars-t-hansen:


lars-t-hansen:
This works as it stands and should be sensibly robust.  The main open question is whether all this information is useful for sysinfo, or some for sysinfo and some for ps.  (For example, fan speed and current clock speeds and memory usage are probably not very useful for sysinfo.)  The other issue that I can't address is AMD - our only AMD machine is no longer working (card drivers too old and incompatible with current OS, and not fixable).  So this is nvidia-only.

lars-t-hansen:
The other issue is that `nvidia-smi -q` is very expensive (takes about 1s to run on an idle workstation with four cards), while extracting just the data we need through a more precise query seems to be much cheaper.  This doesn't matter for `sonar sysinfo` but it will be a concern for `sonar ps`.

lars-t-hansen:
I'll add AMD information too, but I may need more time for that and will file a followup bug - our AMD system has been offline for a while, back just yesterday.  NVIDIA is the high bit anyway, for UiO/NRIS.

lars-t-hansen:
Going to tweak this a little more.

bast:
Thank you!

PR #205, opened 2024-11-13
Fix #185 - add a "slurm" command to extract sacct information

### LINKED ISSUE #185: Roll sacctd into sonar?
There is a utility called sacctd (https://github.com/NAICNO/Jobanalyzer/tree/main/code/sacctd) that runs on one node on a Slurm-managed cluster every hour (say) and extracts information about completed jobs via sacct and reformats and sends this in the usual way to the Jobanalyzer server.  (The benefit of this is that we can retain information longer than the Slurm DB, we can preprocess and cache in more efficient ways, etc.)

Sacctd is a pretty simple program but it becomes one more program to install / update on the cluster.  I'm wondering if it wouldn't be beneficial to roll it into Sonar, in the same way that the sysinfo utility was rolled into Sonar.  One would then run `sonar sacct` to extract this information.

lars-t-hansen:


lars-t-hansen:
Groan, the test output is tied to a time zone, so if the test runs in a different time zone it will fail.

Issue #201, opened 2024-10-25
Collect per-gpu load


lars-t-hansen:
We added the "load" field to detail per-node CPU load per sampling interval (but not per-process because the focus was on node balance and cpu affinity).  And for eg betzy there's full-node allocation anyway so there's just one job running.

But we may want to record per-gpu load and not sum-of-gpus load for a process, if the process runs on more than one gpu.  This could be handled with a new field gpuload= for gpu-using processes and the payload being the same base-45 encoded values as for the load= field.

PR #208, opened 2024-11-15
#201 - Per-gpu load data

lars-t-hansen:
NOTE, this sits on top of #203 and needs to be rebased once that lands!


lars-t-hansen:
Basically done, just need to tidy up a bit.

bast:
Thanks! I will rebase and review.

PR #213, opened 2024-11-25
For #209 - deal with newer rocm-smi output (release_0_12)

lars-t-hansen:
This patch is a bit bigger than it needs to be: it reorganizes the file a little, moving test cases to the end.  Interleaving code and test cases was an experiment (keep tests close to the code) that was a failure.

Issue #209, opened 2024-11-19
rocm-smi output format change breaks sonar


lars-t-hansen:
Old format roughly this:
```
================================= Concise Info =================================                                                                                                                                                           
GPU  Temp (DieEdge)  AvgPwr  SCLK     MCLK    Fan     Perf  PwrCap  VRAM%  GPU%                                                                                                                                                            
0    53.0c           220.0W  1576Mhz  945Mhz  10.98%  auto  220.0W   57%   99%                                                                                                                                                             
1    26.0c           3.0W    852Mhz   167Mhz  9.41%   auto  220.0W    5%   63%                                                                                                                                                             
================================================================================                                                                                                                                                           
```
New format roughly this:
```
=================================================== Concise Info ===================================================
Device  Node  IDs              Temp    Power  Partitions          SCLK    MCLK    Fan    Perf  PwrCap  VRAM%  GPU%
              (DID,     GUID)  (Edge)  (Avg)  (Mem, Compute, ID)                                                    
====================================================================================================================
0       8     0x687f,   18789  29.0°C  3.0W   N/A, N/A, 0         852Mhz  167Mhz  9.41%  auto  220.0W  0%     0%
1       9     0x687f,   28157  25.0°C  3.0W   N/A, N/A, 0         852Mhz  167Mhz  9.41%  auto  220.0W  0%     0%
====================================================================================================================
```
This breaks the parser.

I don't know what the old version was but `rocm-smi --version` reports this for the new version:
```
ROCM-SMI version: 2.3.1+e370dfa
ROCM-SMI-LIB version: 7.3.0
```

lars-t-hansen:
Also see #206 but that's a bigger work item.

lars-t-hansen:
The new format is incompatible with the old since the line below the header is taken as a termination line by the parser.

The current rocm-smi can probably supply us with better and easier-to-parse information by using eg `rocm-smi --showuse --showmemuse --csv`.  I don't know what the old code would have supported.

lars-t-hansen:
The patch did not fix the issue fully, my bad.  If the CSV output succeeds then everything is fine.  Otherwise the code falls through to the older case for parsing text output.  This will hopefully work on older devices but will fail on newer devices.  So if the CSV output fails for any reason - which it does currently due to a config issue on the machine - we'll hit the old code with new output and then we'll crash again.  It's OK to fail, but not OK to crash.

PR #214, opened 2024-11-25
For #209 - Parse new rocm-smi output (main)

lars-t-hansen:
Uplift from release_0_12 (see #213).  This is a little more disciplined, it doesn't move so much code around.


bast:
Thank you!

PR #216, opened 2024-11-27
No bug - very basic GPU logic test cases (release_0_12)

lars-t-hansen:
These tests are run as part of the normal test cycle but will only do something on nodes that have the AMD/NVIDIA cards installed.  Mostly, therefore, actual testing will happen on nodes we control and not on the github runner.

### DIFF
--- a/tests/amd-gpu.sh
+++ b/tests/amd-gpu.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+#
+# Check that `sonar sysinfo` can detect an AMD GPU if it ought to (based on info from the file
+# system).  This test must be run on a node with such a device to have any effect, hence will not be
+# effective in the github runner.
+#
+# Requirement: the `jq` utility.
+
+set -e
+if [[ ! -e /sys/module/amdgpu ]]; then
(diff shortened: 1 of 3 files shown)


lars-t-hansen:
Should be possible to cherry-pick this to main.

bast:
> Should be possible to cherry-pick this to main.

But why not merging? I am asking because I might be missing something.

lars-t-hansen:
In my head merging would be inappropriate since this PR is off of release_0_12, which is not main - and main has diverged from release_0_12.

(Frankly merging is in general a pretty dodgy idea, I'm much more of the opinion that patches always should be cherry-picked onto whatever branches they should go, but as i understand it this is a very contentious topic.)

bast:
Ah I understand now. I guess I could merge the pull request since it won't modify main. And then I could open another pull request from release_0_12 to main to get the patch also to main but maybe it's there already?

lars-t-hansen:
No, I didn't make a PR for it, hence the comment about cherry-picking and the uplift-required tag.  A separate PR seemed needlessly confusing to me, but YMMV.

PR #230, opened 2025-01-13
Fix #229 - gpu probe regression

### LINKED ISSUE #229: Incorrect gpufail=1 after gpu probe updates
Changeset 9d623ce6 introduced a bug: nodes with no GPU will now signal gpufail=1.  This is because the gpu probe only distinguishes between found and not found, and the PS code takes the latter to be an error, but this is wrong and was caused by this patch, which previously would signal gpufail=1 only if the probe returned an Err, it being a Result previously with an empty-array return for no GPUs.

This is a problem on main only, 0.12 is unaffected.

The fact that this bug made it through is a result of not having good test cases for the gpu probes, #217.  We need to be able to run a test suite on hardware that has no GPUs, NVIDIA gpus, AMD gpus and then be able to state expectations for the tests on the various nodes.  It's a little involved and won't work as part of CI, but it's definitely something we want.

lars-t-hansen:
Move the setting of the error flag into an error case, add a test case.

### DIFF
--- a/src/ps.rs
+++ b/src/ps.rs
@@ -342,12 +342,12 @@ fn do_create_snapshot(jobs: &mut dyn jobs::JobManager, opts: &PsOptions, timesta
     let gpu_utilization: Vec<gpu::Process>;
     let mut gpu_info: String = "".to_string();
     match gpu::probe() {
-        None => {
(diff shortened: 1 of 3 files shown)


bast:
Thank you!

PR #231, opened 2025-01-13
Fix #228 - better output logic

### LINKED ISSUE #228: JSON output for sample data
For some consumers, data volume and performance are somewhat secondary concerns and an easily parsed data format is more important.  This is the case for a NAIC collaborator at Simula, who will switch to using Sonar for data collection.  We'll want to emit JSON text from Sonar (under an option) in this case.

lars-t-hansen:
This allows all operations to print csv or json (and in the future, eg a binary format) by abstracting the output logic.  Generally this is a nice improvement since it separates the output logic completely from the rest of the program logic.

TODO here:
- [x] more testing on various hardware
- [x] probably some test cases
- [x] possibly, at the upper level there should be some attention paid to buffering output and flushing the buffer, not sure yet what stdout does (in Go it is unbuffered)
- [x] All JSON packets should have an outermost object that can take an expanding array of named fields, this leads to better extensibility over time (only fields are added, the data format is unchanged), cf #233.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,6 +10,7 @@ mod jobs;
 mod log;
 mod nvidia;
 mod nvidia_nvml;
+mod output;
 mod procfs;
 mod procfsapi;
 mod ps;
@@ -19,6 +20,8 @@ mod sysinfo;
 mod time;
 mod users;
 mod util;
+
+use std::io;
 
 const USAGE_ERROR: i32 = 2; // clap, Python, Go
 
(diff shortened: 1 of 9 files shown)


lars-t-hansen:
stdout is line-buffered (but has no flush method).  By taking a reference to a stdout object and casting it to an io::Write we also gain the ability to call flush on it, which is what we want.

lars-t-hansen:
Converting this back to draft, I'd like to revisit the JSON representation of slurm data, cf #233.

lars-t-hansen:
Rebased and fairly clean, @bast review at will.

bast:
Thank you for the nice abstraction!

PR #236, opened 2025-01-22
Fix #235 - Support aarch64 and non-GPU systems, tidy up gpuapi builds

### LINKED ISSUE #235: Figure out multi-arch gpuapi support
We have some pure aarch64 nodes, and while these do not have GPUs at present we'll still need to link against libsonar-nvidia.a and libsonar-amd.a to be able to compile Sonar at all for the nodes.  So several related issues to solve:

- some kind of stub code for architectures for which we do not have support for a particular gpu
- some way of distinguishing the .a files for different architectures

I sort of expect that for Olivia, we may end up with fat binaries but I don't know that for sure, and if so we will revisit.  For now I just need to make sure that we can compile and run Sonar on our freebio* nodes.

lars-t-hansen:
GPU API fixes:

* Introduce subdirectories for architecture-specific gpuapi libraries
* Make build.rs pick the right subdirectory
* Introduce ifdefs in the gpuapi library sources to allow them to be compiled even if GPU API is unavailable
* Introduce site and node specific build scripts for the libraries

Other aarch64 fixes:

* /proc/cpuinfo is completely different, so parse it properly
* cty::c_char is u8, so clean up the mess around C's char type

Feature system:

* Add Cargo features for amd, nvidia, xpu to complement the stub libraries
* Add a skeleton implementation of XPU that is not supposed to link yet (no libraries)
* Add test cases for feature combinations

Also ran cargo fmt and picked up some random changes that way.

Also cleaned up the text in some test cases.


lars-t-hansen:
The test failure seems to be caused by the return of an old problem wherein the github runner environment is missing some environment variable that the tests depend on.

lars-t-hansen:
Specifically, in #193 we fixed a regex so that user names could contain dashes, now we have a user name with a leading underscore.

bast:
There is a conflict here - should I look at that?

lars-t-hansen:
> There is a conflict here - should I look at that?

If convenient, otherwise I'll deal with in the morning.  If you're happy with the patch in other respects I can merge it myself, LMK.

bast:
Thank you!

PR #251, opened 2025-02-11
Fix #238 - Virtualize all system interfaces to allow for better tests

### LINKED ISSUE #238: Virtualize all the things
Currently system APIs for /proc and the GPUs are virtualized, which greatly aids testing (the GPU bits are coming in with #237).  A few more things remain - eg, access to current time and hostname, there may be more.  We might as well virtualize those too, either putting them into procfs or creating another system virtualization object.

lars-t-hansen:
I think this is done but I want to add a couple patches with new test cases that take advantage of the virtualization before landing it, so making it draft for now.


src/main.rs
```diff
@@ -375,7 +390,8 @@ fn usage(is_error: bool) -> ! {
     let mut stderr = std::io::stderr();
     let out: &mut dyn std::io::Write = if is_error { &mut stderr } else { &mut stdout };
 
-    show_version(out);
+    let system = realsystem::RealSystem::new().freeze();
```
lars-t-hansen, line 393: This is pretty ugly, can't this be a parameter?

src/main.rs
```diff
@@ -142,21 +150,28 @@ fn main() {
                 json: *json,
             };
             if *batchless {
-                let mut jm = batchless::BatchlessJobManager::new();
-                ps::create_snapshot(writer, &mut jm, &opts, &timestamp);
+                let system = realsystem::RealSystem::new()
```
lars-t-hansen, line 153: It's sort of dumb that the system is created so many places, can we fix this?

src/main.rs
lars-t-hansen: The correct answer is that show_version should always print the true version, never the version from the systemapi object.  So just remove this parameterization.

src/main.rs
lars-t-hansen: Yes.

lars-t-hansen:
This is a pretty big change since it moves a lot of things around, but there should be basically no functional change.  The gist is that there's a SystemAPI, a GpuAPI, and a ProcfsAPI that allow us to mock system calls, GPU probing, and /proc reading.  The ProcfsAPI is not new, it's just cleaned up - things that were on that object but didn't belong there are moved to the SystemAPI, along with many other things.

To take advantage of this, there are two implementations of those APIs: the "real" implementations going to the OS, and the "mock" implementations that can be instantiated with controlled data.  Using the latter we can test end-to-end functionality with controlled results.  I've introduced a couple of tests already on the second patch here.  More should be added later, now that we can.

bast:
Thanks so much! This is an important improvement.

Issue #252, opened 2025-02-11
Revamp the JSON format


lars-t-hansen:
The JSON format should be redone to follow the json:api specification, and we should consolidate fields with slurm-monitor when possible.  This is a good time to do so since there are not currently any users of our JSON format apart from our test cases.

PR #255, opened 2025-02-26
Fix #147 - Daemon mode, built-in Kafka exfiltration

### LINKED ISSUE #147: Sonar "daemon" mode with Kafka exfiltration
Instead of piping the output from Sonar to curl, what is now `exfiltrate` in Jobanalyzer could be integrated wholesale into Sonar, using eg [reqwest](https://docs.rs/reqwest/latest/reqwest/).  On the pro side, we don't depend on curl and we don't have to run a pipeline and two programs; reduces both complexity and load.  On the con side, more complexity internal to Sonar.

I envision a `--target` switch specifying the target URL, optionally - otherwise output goes to stdout.  We'd also need a switch or two to specify secrets: root certificates for the server and upload permissions for the HTTP connection.  This is as in exfiltrate.

(This is related to #143 clearly since that is a bit of exfiltrate that is moving into Sonar.) 

Mostly a matter of taste?  No sense in doing this until we're sure it's the right thing to do, but wanted to examine the possibility.

lars-t-hansen:
THIS IS OBSOLETE, REPLACED BY https://github.com/NordicHPC/sonar/pull/280.

This is working, pretty much.  It may be that this is just the first of several patches!  The initial goal here is to get to a working prototype, possibly without all the required features, so that we can play with the pipeline.

For sure important for a working prototype:

- [x] filtering logic for PENDING / RUNNING data to keep the volume sane
- [x] api token for authentication / authorization
- [ ] some kind of secrecy, ideally TLS

Probably important pretty soon and maybe something we have to do here:

- [x] synchronizing alarms to the cadence to avoid clock drift
- [x] garbage collecting job records in the filter

Probably not important yet

- verify that the delta coding of slurm data is worth the bother, see comment further down
- randomized sending times, modulo what Kafka does
- control pileup of unsent messages
- connection shutdown
- misc error handling
- how to handle control messages correctly and efficiency and without local storage, also when multiple clients receive the same control message
- implement lock file
- better signal handling
- reload config
- pause / restart remote control?
- improve cadence computation (90m is a thing)
- old -> new translation code
- for cadences of > 1h, they should probably be synchronized to local time, not utc.  Consider a cadence of 4h: now it runs at 02:00, 06:00, etc (b/c Oslo is DST+02:00 at the moment).  But activity on a system is most tied to local time, so sending at 00:00, 04:00, etc local time would make most sense, and this is what cron would do
- test cases


lars-t-hansen:
Somebody broke cargo; it keeps updating the version in Cargo.lock to v4.  This breaks everything, and the Internet is on the case.  Meanwhile, if this should happen again, manually change it back to 3, for the stuff we have no other changes are needed.

We need Rust v1.67.0 or newer now because flate2, used by the Kafka library, requires it (verified with cargo msrv).  When it becomes time to land this we can discuss options.  Rust 1.67 is more than two years old as I write this so it's probably not a catastrophy.

README.md
```diff
@@ -91,7 +91,8 @@ With the branches come some additional rules for how to move patches around:
 
 At the time of writing we require:
 - 2021 edition of Rust
-- Rust 1.59.0, released 2022-02-24 (can be found with `cargo msrv find`)
+- Rust 1.67.0, released 2023-01-26 (can be found with `cargo msrv find`).  Probably 1.67.1 is the
```
lars-t-hansen, line 94: See comment about MSRV in the test case.

bast:
IMO no problem to require Rust 1.67 or newer.

lars-t-hansen:
Queue rebased to sit on top of the queue for #252.

lars-t-hansen:
Re the delta coding of slurm jobs data: with a sufficiently high cadence of transmission and with the inclusion of pending and running jobs, the data volume on a large cluster may be very significant.  For a 5m cadence on Fox the delta coding cuts the volume 10x-70x (based on too few data points but it's credible), and then that can be further compressed using standard compression.  So at first blush this is valuable.  However, it introduces complexity on both the sending side and the receiving side, so it had better be worth it in terms of data savings, *after* we apply standard data compression to both the normal and the delta-compressed streams.  Assessing this properly will require setting up some tests on all the large clusters, and measuring the compressed data sizes of both streams (they don't actually have to be sent).  The cadence is *not* irrelevant for this, as a longer interval between probes will yield larger deltas.  We need to pick a credible cadence.  We could run experiments for multiple cadences, say, 1m, 5m, 15m, 30m, 1h.  This is a fair amount of work.  So the *first* order of business is to assess what the compressed volume for non-delta-encoded data would be from these clusters, to even motivate going further.

Issue #257, opened 2025-03-07
Improve documentation


lars-t-hansen:
Currently we have one long run-on README.md file.  Let's improve this.

For one thing, output formats / fields could be documented separately.  With my w-daemon branch I'm adding machine-executable documentation of the output format (aka data structures and a very simple-to-understand parser), which is one thing.  Another thing is that for NAIC we are documenting the sonar output format and field semantics extensively, currently in a GDoc but this can and should be moved to markdown in this repo.

For another thing, the changelog could be broken out.

There must be more.

bast:
We could move documentation to Sphinx - the same engine we use within NRIS. Or it could be anything else but it's important for me that the documentation stays in this repo, ideally in Markdown format and we use something that generates HTML.

lars-t-hansen:
y, keeping it in this repo is a hard requirement imo.  just linked markdown is fine with me, it doesn't need to be fancy.  happy to discuss.

lars-t-hansen:
I'm going to remove this from the MVP again.  I've done some cleanup today and apart from deployment docs, which are covered by #346, I think what remains here is long-term work to put the docs into a different format.

lars-t-hansen:
Docs were improved enough as part of the process to get Sonar onto Saga in 2026Q1.

PR #258, opened 2025-03-10
For #257 - Reorganize + flesh out doc, add executable spec

lars-t-hansen:
@bast, lmk what you think about this.  This is not "the" document cleanup but a start.  It splits the data format spec and the changelog out and places them in doc/.  Then it elaborates a lot on the data format spec, there were many things missing.  Finally it provides a formal-ish spec of the data format in the form of type definitions and some code to parse text into those types; I have used Go since it's high-level and easy for this type of thing (and allows types to be defined at all, unlike Python, say).

This is in anticipation of some major developments that are right around the corner, namely, a new JSON data format, to be specified separately (and backed by similar code), and a "daemon mode" that keeps Sonar memory-resident and exfiltrates data via Kafka.  I figured it would be easier to factor this change and not complicate the subsequent patches with it.


lars-t-hansen:
While I think this is OK as it is since we'll be replacing the old format with a new format anyway, it could be noted somewhere that the Slurm AllocTRES field has a lot of substructure.  In the current format we're just dumping it verbatim as a string, and this is probably OK - the client's just going to store it in a text file anyway.  With the new format, we might discuss whether Sonar decodes AllocTRES or if we push that to the client.

bast:
This is a good step. Thank you! Having the markdown a bit more bite-sized under doc/ will make it easier to brows and later also super easy to bolt Sphinx on top of it to have it rendered in a nice HTML site.

lars-t-hansen:
@bast, this became a little over-elaborate but is probably ready for final review.  The main point is still the doc reorg, nothing has changed there.  The code and data in util/ are going to be highly useful going forward but has ballooned a bit because I added test code and test data.  More could be done, and I think that over time should be done, because this is the best way to keep sonar completely honest about what it emits, but now is not the time for more than what is already here (and it's already quite a bit).

bast:
Thank you!

PR #259, opened 2025-03-13
Fix #182 - remove --batchless, emit jobid epoch

### LINKED ISSUE #182: Get rid of `--batchless` option
On slurm systems, it will sometimes happen that a node is taken out of the slurm pool and turned into an interactive node, this seems to happen with some regularity on Fox.  The admins are unlikely to remember to change the options to sonar in this case, or to change them back when the node goes back into the pool: non-slurm nodes need the `--batchless` option, slurm nodes run with `--rollup`.  This turns into a mess on the backend where we have inappropriate job IDs for all the jobs from an interactive node sonar'd without `--batchless` (https://github.com/NAICNO/Jobanalyzer/issues/534)

Probably we should try to get rid of `--batchless` and just infer it.

On a slurm system, we will be able to find the slurm job ID in /proc/pid/cgroup (we should be a little more fastidious about what we match when we search that file).  If the node is interactive, those data should not be there, and in that case, we should be able to fall back on using the process group, as we do for non-slurm systems.

One real complication is that a lot of processes on a slurm system will not have a slurm ID, because they are not slurm jobs, just normal processes.  In that case, we currently use 0 for the job ID, as we must; no PID will be safe against confusion with the slurm job ID.  This is unlike on batchless systems where we always use the pgrp except in corner cases (where we use 0).  On a given system we will therefore need to compute slurm and pgrp job IDs for all the processes, and if we find a nonzero slurm ID for any process then we use the slurm IDs (and zeroes), and otherwise we use the pgrp IDs.

There will be a bit of surgery required for this, the "jobmanager" abstraction will probably disappear, and the data structures and printing will be embellished.  We should be sure to leave copious comments.

Now about `--rollup`.  If present, it should only have effect if we have Slurm job IDs (or other batch system IDs I guess).  This would require a bit more surgery still.

lars-t-hansen:
I opted to allow the old --batchless argument for now, but without any effect at all.  Will remove it when I'm sure all scripts have been updated.


src/jobsapi.rs
```diff
@@ -21,7 +23,34 @@ impl NoJobManager {
...
+        AnyJobManager {
+            force_slurm,
+        }
+    }
+}
+
+impl JobManager for AnyJobManager {
+    fn job_id_from_pid(&self, pid: usize, processes: &HashMap<usize, procfs::Process>) -> (usize,bool) {
+        if let Some(slurm_job_id) = slurm::get_job_id(pid) {
+            match slurm_job_id.trim().parse::<usize>() {
```
lars-t-hansen, line 46: I guess really the number parsing should move into slurm::get_job_id().

src/slurm.rs
```diff
@@ -1,22 +1,11 @@
 // Abstraction of jobsapi::JobManager for SLURM.
+//
+// TODO: Shouldn't this be virtualized through procfs?
```
lars-t-hansen, line 3: I should do this, and write a test, b/c we have no good tests for this.

src/main.rs
```diff
@@ -244,22 +239,17 @@ fn command_line() -> Commands {
                 }
 
                 #[cfg(debug_assertions)]
-                let allow_incompatible = std::env::var("SONARTEST_ROLLUP").is_ok();
+                let force_slurm = std::env::var("SONARTEST_ROLLUP").is_ok();
```
lars-t-hansen, line 242: These don't belong here, but before invoking ps, since they are no longer relevant to argument parsing.

src/systemapi.rs
```diff
@@ -6,15 +6,19 @@ use std::fs;
...
+    // The `get_` methods always return the same values for every call.  `btime` is the boot time
+    // minus BTIME_BASE.
     fn get_version(&self) -> String;
     fn get_timestamp(&self) -> String;
     fn get_hostname(&self) -> String;
     fn get_clock_ticks_per_sec(&self) -> usize;
     fn get_page_size_in_kib(&self) -> usize;
     fn get_now_in_secs_since_epoch(&self) -> u64;
     fn get_pid(&self) -> u32;
+    fn get_btime(&self) -> u64;
```
lars-t-hansen, line 21: This is just a terrible name for it.

bast:
Thank you!

PR #261, opened 2025-03-19
Fix #252 - new JSON format

### LINKED ISSUE #252: Revamp the JSON format

lars-t-hansen:
This is pretty massive and is structured as a queue that should be reviewed from top to bottom, starting with "Sundry cleanup".  In addition to introducing the new JSON format it adds a number of fields that will be required by that format.  The only really complicated change here is the change to `sonar ps`, which should probably be considered a rewrite.

Consolidated to-do list:

- [x] the cpu_util field
- [x] field names for all the slurm data are subject to change, but may have to be a subsequent patch
- [x] some documentation
- [x] some test cases
  - [x] as noted below
  - [x] json syntax tests for slurm
  - [x] (more)
- [x] (maybe) --json should probably cause --cluster to be a required option

Again, the evolving spec is [in an evolving NRIS-space gdoc](https://docs.google.com/document/d/1rpecBCIpOgyO2JBD8N9RqtIn0wM0Jkn1COSV3Fy7860/edit?usp=sharing).


lars-t-hansen:
I'll be reviewing and testing and probably making some small tweaks, but at this point, barring major comments, it may be easier to collect any further tweaks in a separate patch at the end of this queue.

util/formats/newfmt/types.go
```diff
@@ -0,0 +1,217 @@
...
+type SysinfoAttributes struct {
+	Time        string                   `json:"time"`
+	Cluster     string                   `json:"cluster"`
+	Node        string                   `json:"node"`
+	OsName      string                   `json:"os_name"`
+	OsRelease   string                   `json:"os_release"`
+	Cores       []SysinfoCoreModel       `json:"cores"`
+	Memory      uint64                   `json:"memory"`
+	Description string                   `json:"description"`
+	TopoSVG     string                   `json:"topo_svg"` // Base64-encodes SVG
```
lars-t-hansen, line 55: "encoded"

util/formats/newfmt_test.go
```diff
@@ -0,0 +1,265 @@
...
+	// Here I want to:
+	// - run `sonar sysinfo` once
+	// - parse the output in strict mode
+	// This ensures that:
+	// - all emitted, known fields have the right types
+	// - no unknown fields are emitted
+	// Do this on enough machines and we'll have a decent test of whether Sonar works in practice.
+}
+
+// TODO: Samples
```
lars-t-hansen, line 102: Do it, then.

util/formats/newfmt_test.go
```diff
@@ -0,0 +1,265 @@
...
+	err := newfmt.ConsumeJSONSysinfo(f, true, func(info *newfmt.SysinfoEnvelope) {})
+	assert(t, err != nil && strings.Index(err.Error(), "unknown field") != -1, "Unknown field #1 msg")
+
+	f = strings.NewReader(`{"meta":{"zappa":"hello"}}`)
+	err = newfmt.ConsumeJSONSysinfo(f, true, func(info *newfmt.SysinfoEnvelope) {})
+	assert(t, err != nil && strings.Index(err.Error(), "unknown field") != -1, "Unknown field #2 msg")
+}
+
+func TestNewJSONSysinfoActual(t *testing.T) {
+	// Here I want to:
```
lars-t-hansen, line 93: Looks unfinished.

src/gpuapi.rs
```diff
@@ -1,13 +1,28 @@
...
 
-// Per-sample process information, across cards.  The GPU layer can report a single datum for a
-// process across multiple cards, or multiple data breaking down the process per card even if the
-// process is running on multiple cards.
+use crate::ps::UserTable;
+
+#[derive(PartialEq, Eq, Hash, Default, Clone, Debug)]
+pub struct GpuName {
+    pub index: i32,             // May change at boot time
+    pub uuid: String,           // Forever immutable
```
lars-t-hansen, line 8: Must do: the uuid needs to be on a firmer footing because we use it for everything.  The uuid MUST NOT under any circumstances be confused with some other device, but if a good uuid is not available then it is acceptable for a card to be seen to have multiple UUIDs over time.  For example, our AMD cards do not have a reliable UUID.  It would be acceptable to construct one from, say, hostname:boot_time:card_index (where hostname is as fully qualified as possible, ideally both cluster and node name) so long as we can assume the card index does not change while the machine is up; bus address is an alternative and may be better.  Anyway, probably the lower-level GPU layer should return "" if it can't provide a reliable one, and then we can construct a uuid in a common manner at the gpuapi layer.

src/sysinfo_test.rs
```diff
@@ -18,7 +18,7 @@ pub fn sysinfo_error_test() {
 #[test]
```
lars-t-hansen, line 9: Does this need to be ifdef x86_64?

util/formats/newfmt/types.go
```diff
@@ -0,0 +1,217 @@
+// New Sonar JSON data format (we no longer support any kind of CSV variant).  MetadataObject and
+// ErrorObject are shared between the various data types, everything else is specific to the data
+// type and has a name that clearly indicates that.  The top-level object for all data types is the
+// "...Envelope" object.  Within the envelopes, `Data` and `Errors` are exclusive of each other.
+//
+// Some fields present in the older Sonar data are no longer here, having been deemed redundant or
+// obsolete.  Some are here in a different form.  Therefore, while old and new data are broadly
+// compatible, there may be some minor problems translating between them.
+//
+// The exact semantics of fields are still defined in the spec in doc/DATA-FORMATS.md.
```
lars-t-hansen, line 10: This file needs to be extracted from the gdoc spec and added.  Actually an argument could be made that it should be a machine-readable spec and that the present file should be generated from that spec.

util/formats/newfmt/types.go
```diff
@@ -0,0 +1,217 @@
...
+type SampleData struct {
+	Type       string           `json:"type"`
+	Attributes SampleAttributes `json:"attributes"`
+}
+
+type SampleAttributes struct {
+	Time    string        `json:"time"`
+	Cluster string        `json:"cluster"`
+	Node    string        `json:"node"`
+	Attrs   []KVPair      `json:"attrs"`
```
lars-t-hansen, line 106: I don't like "attrs" both because it is an abbreviation and because it is confusing vis-a-vis "attributes", of which it is part.  Could this be "properties"?  Do we even want it here?  In slurm-monitor this carries eg the Python version.  Is that a relevant attribute of the sample data, as here, or is it metadata?  It depends.  If we want to encode the command line switches, which affect the data, it's definitely relevant here.  Is it maybe "environment"? "invocation"?

util/formats/newfmt/types.go
```diff
@@ -0,0 +1,217 @@
...
+	Gpus       []SampleGpu `json:"gpus"`
+	UsedMemory uint64      `json:"used"`
+}
+
+type SampleCpu = uint64
+
+type SampleGpu struct {
+	Index      uint64 `json:"index"`
+	UUID       string `json:"uuid"`
+	Bad        bool   `json:"bad"`
```
lars-t-hansen, line 123: "bad" is a bad name.  And currently we don't emit this at all, which is some kind of information loss, b/c earlier we could emit "unknown" for the gpuset in a situation where at least one gpu was borked.

util/formats/newfmt/types.go
```diff
@@ -0,0 +1,217 @@
...
+	Fan        uint64 `json:"fan"`
+	Mode       string `json:"mode"`
+	Perf       int64  `json:"perf"`
```
lars-t-hansen, line 126: "perf" is a bad name b/c abbreviation and unspecific.  Probably "performance_mode", and "mode" should be "compute_mode".

util/formats/newfmt/types.go
```diff
@@ -0,0 +1,217 @@
...
+	Memory     uint64 `json:"memory"`
+	CEUtil     uint64 `json:"ce_util"`
+	MemUtil    uint64 `json:"memory_util"`
+	Temp       int64  `json:"temp"`
```
lars-t-hansen, line 130: "temp" -> "temperature".

util/formats/newfmt/types.go
```diff
@@ -0,0 +1,217 @@
...
+	Processes []SampleProcess `json:"processes"`
+}
+
+type SampleProcess struct {
+	Resident  uint64             `json:"resident"`
+	Virtual   uint64             `json:"virtual"`
+	Cmd       string             `json:"cmd"`
+	Pid       uint64             `json:"pid"`
+	ParentPid uint64             `json:"ppid"`
+	CpuAvg    float64            `json:"cpu_avg"`
```
lars-t-hansen, line 149: These four fields - two here, two in gpu data - need not be floating point, they could be uint.  They are percentage points, not fractions of 1.

util/formats/newfmt/types.go
```diff
@@ -0,0 +1,217 @@
...
+
+type JobsData struct {
+	Type       string         `json:"type"`
+	Attributes JobsAttributes `json:"attributes"`
+}
+
+type JobsAttributes struct {
+	Time      string     `json:"time"`
+	Cluster   string     `json:"cluster"`
+	// There can be other types of jobs, there will be other fields for them here, and the decoder
```
lars-t-hansen, line 179: "There can be" -> "There can eventually be"

util/formats/newfmt/types.go
```diff
@@ -0,0 +1,217 @@
...
+	Power      uint64 `json:"power"`
+	PowerLimit uint64 `json:"power_limit"`
+	CEClock    uint64 `json:"ce_clock"`
+	MemClock   uint64 `json:"memory_clock"`
+}
+
+type SampleJob struct {
+	Job       uint64          `json:"job"`
+	User      string          `json:"user"`
+	Processes []SampleProcess `json:"processes"`
```
lars-t-hansen, line 140: FIXME!  "epoch" is missing and is also not emitted.

util/formats/newfmt_test.go
```diff
@@ -0,0 +1,265 @@
+// The purpose of this test:
```
lars-t-hansen, line 1: Add a purpose: check correspondence between Rust emitter and newfmt/types.go.

util/formats/newfmt_test.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
+	// There are three records: one for ml1 with GPUs, one for c1-6.fox without, one error
+	var iter int
+	err = newfmt.ConsumeJSONSysinfo(f, false, func(info *newfmt.SysinfoEnvelope) {
+		switch iter {
+		case 0:
+			assert(t, info.Meta.Producer == "sonar", "#0 producer")
```
lars-t-hansen, line 32: TODO: Check correspondence between this and current definition.

src/sysinfo.rs
```diff
@@ -5,33 +5,187 @@ use crate::systemapi;
...
+    sysinfo.push_s("version", system.get_version());
+    sysinfo.push_s("timestamp", system.get_timestamp());
+    sysinfo.push_s("hostname", system.get_hostname());
+    sysinfo.push_s("error", error);
+    sysinfo
+}
+//-oldnames
+
+// Old and new formats layout the cards the same way but use different field names in a number of
+// cases.
```
lars-t-hansen, line 106: This code is brittle wrt the test that greps for strings and compares them to known fields.  It will work fine so long as the new names are used in the "if" part of the tests below and nothing goes over a line break.  May note that here.  Long-term the old-fmt code will go away I guess.

src/sysinfo.rs
```diff
@@ -5,33 +5,187 @@ use crate::systemapi;
...
+                   *max_mem_clock_mhz as i64);
+        gpu_info.push_o(gpu);
+    }
+    gpu_info
 }
 
 const GIB: usize = 1024 * 1024 * 1024;
 
-fn try_compute_sysinfo(system: &dyn systemapi::SystemAPI) -> Result<output::Object, String> {
+struct CoreInfo {
```
lars-t-hansen, line 155: This is redundant now, we could use procfs::CoreInfo instead.

src/ps_newfmt.rs
```diff
@@ -0,0 +1,179 @@
...
+        }
+        sstate.push_a("cpus", cpu_load);
+        if let Some(gpu_samples) = &c.gpu_samples {
+            let mut gpu_load = output::Array::new();
+            for v in gpu_samples {
+                gpu_load.push_o(format_newfmt_gpu_sample(v));
+            }
+            sstate.push_a("gpus", gpu_load);
+        }
+        // sstate.push_u("used")  FIXME
```
lars-t-hansen, line 31: We don't even compute this, I think?  Where does it come from?

src/ps_newfmt.rs
```diff
@@ -0,0 +1,179 @@
...
+    for k in zeroes {
+        let j = &c.process_samples[k];
+        jobs.push_o(format_newfmt_job(0, &j.user, &vec![k], &c.process_samples));
+    }
+    for ((user,id),ixs) in collected {
+        jobs.push_o(format_newfmt_job(id, user, &ixs, &c.process_samples));
+    }
+    // attr.push_s("key") // FIXME
+    // attr.push_s("value") // FIXME
+    // attrs.push_a("attrs") // FIXME
```
lars-t-hansen, line 60: See earlier comment about why this is a bad name.

src/ps_newfmt.rs
```diff
@@ -0,0 +1,179 @@
...
+
+fn format_newfmt_gpu_sample(c: &gpuapi::CardState) -> output::Object {
+    let mut s = output::Object::new();
+    if c.device.index != 0 {
+        s.push_i("index", c.device.index as i64);
+    }
+    if c.device.uuid != "" {
+        s.push_s("uuid", c.device.uuid.clone());
+    }
+    // s.push_b("bad") // FIXME
```
lars-t-hansen, line 75: It's a bad name, and we used to compute this, so why not now?

src/ps_newfmt.rs
```diff
@@ -0,0 +1,179 @@
...
+    samples: &[ProcInfo],
+) -> output::Object {
+    let mut job = output::Object::new();
+    job.push_u("job", id as u64);
+    job.push_s("user", user.to_string());
+    let mut procs = output::Array::new();
+    for ix in ixs {
+        procs.push_o(format_newfmt_sample(&samples[*ix]));
+    }
+    job.push_a("processes", procs);
```
lars-t-hansen, line 125: TODO: epoch.

src/ps_newfmt.rs
```diff
@@ -0,0 +1,179 @@
...
+        // and sonar has no history.
+        fields.push_u("pid", proc_info.pid as u64);
+    }
+    if proc_info.ppid != 0 {
+        fields.push_u("ppid", proc_info.ppid as u64);
+    }
+    if proc_info.cpu_percentage != 0.0 {
+        fields.push_f("cpu_avg", three_places(proc_info.cpu_percentage));
+    }
+    // fields.push_u("cpu_util") // FIXME
```
lars-t-hansen, line 151: Probably need cpu_util.

src/ps.rs
```diff
@@ -250,444 +132,450 @@ fn do_create_snapshot(
...
-            if print_params.flat_data {
-                output::Value::A(output::Array::from_vec(vec![output::Value::O(hb)]))
+            if opts.new_json {
+                let mut envelope = output::newfmt_envelope(system);
+                envelope.push_a("errors", output::newfmt_one_error(system, error));
+                output::write_json(writer, &output::Value::O(envelope));
             } else {
-                output::Value::O(hb)
+                let mut hb = make_oldfmt_heartbeat(system);
+                hb.push_s("error", error);
```
lars-t-hansen, line 160: Hm, this does not need a +oldnames annotation, which suggests that test coverage is not as good as it needs to be (we're not checking ps.rs).

bast:
Thanks! I need to carefully read through these.

src/ps.rs
```diff
@@ -2,135 +2,17 @@
 #![allow(clippy::too_many_arguments)]
```
lars-t-hansen, line 2: These pragmas might not apply any more.

src/ps.rs
```diff
@@ -2,135 +2,17 @@
...
-                gpu_mem_percentage,
-                gpu_mem_size_kib,
-                gpu_status: GpuStatus::Ok,
-            }
-        });
-}
-
 #[derive(Default)]
 pub struct PsOptions {
     pub rollup: bool,
```
lars-t-hansen, line 18: PsOptions::always_print_something is only ever true.

src/ps.rs
```diff
@@ -250,444 +132,450 @@ fn do_create_snapshot(
...
+    device: &gpuapi::GpuName,
+    gpu_util: u32,
+    mem_util: u32,
+    mem_size: u64,
+) -> GpuProcInfos {
+    let mut h = HashMap::new();
+    h.insert(device.clone(), GpuProcInfo{
+        device: device.clone(),
+        gpu_util,
+        gpu_mem: mem_size as u64,
```
lars-t-hansen, line 453: The cast is not necessary.

src/ps.rs
```diff
@@ -250,444 +132,450 @@ fn do_create_snapshot(
...
+    device: &gpuapi::GpuName,
+    gpu_util: u32,
+    mem_util: u32,
+    mem_size: u64,
+) {
+    gpus.entry(device.clone())
+        .and_modify(|e| {
+            e.gpu_util += gpu_util;
+            e.gpu_mem_util += mem_util;
+            e.gpu_mem += mem_size as u64;
```
lars-t-hansen, line 470: The cast is not necessary (also below).

lars-t-hansen:
> Thanks! I need to carefully read through these.

Yes, it's a big job :-(  The `sonar ps` patch might be most easily reviewed by ignoring the diff and instead just looking at the final files on my work branch.  The others are reasonably clean, fortunately.

src/gpuapi.rs
lars-t-hansen: Resolved by a tweak in the AMD card code + documentation that kicks the can down the road for other cards.

src/ps.rs
lars-t-hansen: Will be removed in the final ps patch.

src/sysinfo_test.rs
lars-t-hansen: yes.

util/formats/newfmt/types.go
```diff
@@ -0,0 +1,217 @@
...
+//
+// Some fields present in the older Sonar data are no longer here, having been deemed redundant or
+// obsolete.  Some are here in a different form.  Therefore, while old and new data are broadly
+// compatible, there may be some minor problems translating between them.
+//
+// The exact semantics of fields are still defined in the spec in doc/DATA-FORMATS.md.
+
+package newfmt
+
+type MetadataObject struct {
```
lars-t-hansen, line 14: The spec adds a "format" field here.

util/formats/newfmt/types.go
lars-t-hansen: In the spec this has moved to metadata, which makes some type of sense since eg the producer name and version are there, and the data envisioned for "attrs" is that type of data.

util/formats/newfmt/types.go
lars-t-hansen: That can't happen until final review however, so we might do it as a separate Issue.

util/formats/newfmt/types.go
lars-t-hansen: I will not do this now.  They have been float forever, and they can change to integer later without affecting any clients that consume untyped data a la JSON (typed data - maybe not so much).

util/formats/newfmt/types.go
lars-t-hansen: Once upon a time, this was an error code that could have more than two values...

src/ps_newfmt.rs
lars-t-hansen: Will be fixed as part of the spec patch.

src/ps_newfmt.rs
lars-t-hansen: Was fixed.

src/ps_newfmt.rs
lars-t-hansen: Will be fixed as part of the spec patch.

src/ps_newfmt.rs
lars-t-hansen: The name is also not good.  The spec has now renamed it "used_memory".

lars-t-hansen:
Aargh, misled by `cargo clippy`.

lars-t-hansen:
What remains after my own review is probably just the `cpu_util` field and then some documentation and test cases.

bast:
Sorry - maybe I made it conflict by merging the PRs out of order.

lars-t-hansen:
> Sorry - maybe I made it conflict by merging the PRs out of order.

Can't be helped.  I'll rebase everything.

PR #266, opened 2025-04-02
Fix #263 - NVIDIA library loading and versioning

### LINKED ISSUE #263: Allow the use of older NVIDIA APIs
The current nvidia code is written against "recent" cuda headers and libraries, using eg `nvmlDeviceGetComputeRunningProcesses_v3()`.  But sometimes nodes have older cuda, one node I have access to has cuda 470 instead of our normal range 535-560.

Also it could look like these older cuda installs place the library in /lib/x86_64-linux-gnu/ rather than in /usr/lib64 (needs proper investigation), so we'll have to deal with that too.

lars-t-hansen:
This tries to systematically support versioned APIs, and apart from one NVIDIA snafu described in a comment in the code this is not too hard.

Also tries a multitude of different paths for library loading.


bast:
Thank you!

PR #268, opened 2025-04-05
replace lazy_static by once_cell; closes #267

### LINKED ISSUE #267: Replace lazy_static by once_cell
I did that also in another project of mine.

Context: https://github.com/rust-lang-nursery/lazy-static.rs/issues/214

bast:


bast:
@lars-t-hansen for info

bast:
Test fails. It looks like I am using a newer version on my computer?

lars-t-hansen:
The Rust people introduced a bug in Cargo - it pointlessly upgrades the version to 4.  I thought a fix was forthcoming but I've seen nothing yet.  You need to manually downgrade the version to 3.

lars-t-hansen:
once_cell on crates.io says it requires 1.65 so this changes the msrv for sonar (which is fine, I have patches coming taking us beyond that anyhow).  But the doc also says "If only the std, alloc, or race features are enabled, MSRV will be updated conservatively, supporting at least latest 8 versions of the compiler." which means really that currently it could be requiring about 1.77.

bast:
I will downgrade Cargo to 3. And locally I use cargo 1.85.1 so I did not notice and good point that we need to be a bit careful about which version we require.

src/slurmjobs.rs
```diff
@@ -5,36 +5,35 @@ use crate::output;
...
 use std::collections::HashSet;
 use std::io;
 
-lazy_static! {
-    // The job states we are interested in collecting information about, notably PENDING and RUNNING
-    // are not here by default but will be added if the --deluge option is given.
+// The job states we are interested in collecting information about, notably PENDING and RUNNING
+// are not here by default but will be added if the --deluge option is given.
 
-    //+ignore-strings
```
lars-t-hansen, line 19: It would have been necessary to keep the ignore-strings comments previously (they are for a test case), but with the latest patch we landed it's fine to remove them.

lars-t-hansen (approved):
I approve in principle, I guess we just need to get the versions right.

src/slurmjobs.rs
bast: Ah! Good to know. Thanks.

lars-t-hansen:
I'm going to poach this and get it landed.

lars-t-hansen:
Superseded by another patch I just landed.

PR #272, opened 2025-04-09
apply "cargo fmt"

bast:
I kept old formatting in src/sysinfo.rs where it might help reading and comparing lines with old formatting.

### DIFF
--- a/src/slurmjobs.rs
+++ b/src/slurmjobs.rs
@@ -1,10 +1,10 @@
 // Run sacct, extract output and reformat as CSV or JSON on stdout.
 
+use crate::json_tags::*;
 use crate::nodelist;
 use crate::output;
 use crate::systemapi;
 use crate::time;
-use crate::json_tags::*;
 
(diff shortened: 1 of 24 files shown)


src/slurmjobs.rs
```diff
@@ -415,77 +437,83 @@ fn push_uint(obj: &mut output::Object, name: &str, val: &str) {
...
-                n
-            } else {
-                return;
-            }
-        } else {
-            0
-        };
-    let minutes =
-        if let Ok(n) = elts[0].parse::<u64>() {
+    let days = if let Some((dd, rest)) = val.split_once('-') {
```
lars-t-hansen, line 462: Obviously I'm not going to argue with fmt, so this is just me airing a grievance, but this type of decision is just fundamentally wrong, since the nested if will appear at lower nesting levels (visually) than the outer if.

lars-t-hansen (approved):

PR #273, opened 2025-04-10
gpu: fix extraction of min / max power limit

2maz:
Seems gpu min max power_limits had been accidentally swapped


lars-t-hansen (approved):
Oops!

PR #277, opened 2025-04-16
applied cargo clippy --fix --allow-dirty --allow-staged

bast:
- ignored some changes back if it helps the reading/understanding of reading
- did not fix one error found by clippy which I was unsure how to test
- this might break the CI since I am locally using a later version of cargo so we might need to unstage some changes to make it pass


bast:
I made it draft so that I can "downgrade" some of the code to make it pass below the minimum Rust version test.

bast:
Feel free to squash the commits.

src/nodelist.rs
```diff
@@ -48,7 +48,7 @@ struct NodelistParser<'a> {
     i: usize,
 }
 
-impl<'a> NodelistParser<'a> {
+impl NodelistParser<'_> {
```
lars-t-hansen, line 51: TIL!  I ran into this clippy remark myself and was stumped at how to fix it.  I've no idea why this fix makes sense, but I'm OK with it.

src/output.rs
```diff
@@ -420,10 +420,10 @@ pub fn newfmt_envelope(
     if crate::OUTPUT_FORMAT != 0 {
         meta.push_u(METADATA_OBJECT_FORMAT, crate::OUTPUT_FORMAT)
     }
-    if token != "" {
+    if !token.is_empty() {
```
lars-t-hansen, line 423: As a general remark, I think Clippy is just plain wrong here: this is not an improvement.  Maybe if there were an is_not_empty() method we could use, but not when the condition is being split up like this.

src/output.rs
```diff
@@ -420,10 +420,10 @@ pub fn newfmt_envelope(
...
         meta.push_s(METADATA_OBJECT_TOKEN, token)
     }
-    if attrs.len() > 0 {
+    if !attrs.is_empty() {
```
lars-t-hansen, line 426: Same remark here.

lars-t-hansen (approved):

src/output.rs
bast: Would you prefer me to un-stage these kind of changes from this PR?

src/output.rs
bast: Would you prefer me to un-stage these kind of changes from this PR?

lars-t-hansen: Probably not - I recognize a lost cause when I see one, I hope.  I guess I needed to vent about it - IMO it's an annoying amount of (non-ergonomic) drama caused by somebody introducing a second way to do something that was already possible to do.  Maybe there's a justification for this in the realm of generics.  Anyway, let's just do it the way you've done it and be done with it.

bast: Generally, I find clippy useful and it helped me to improve code generally but we don't need to be clippy "certified" from my perspective. I see how "if something is not empty" reads clunky and possibly does not help the understanding.

bast: I undid all the changes that introduced .is_empty() conditionals.

lars-t-hansen (approved):
Yes, looks good!

src/gpuapi.rs
```diff
@@ -1,5 +1,5 @@
+#![allow(clippy::upper_case_acronyms)]
```
lars-t-hansen, line 1: I will take an action item to deal with this, I think renaming the trait as "Gpu" ought to work OK.

PR #279, opened 2025-04-23
Fix #260 - tidy up GPU API

### LINKED ISSUE #260: Clean up gpuapi further
As part of #252 I clean up the gpuapi some but it is far from enough.  The types in the gpuapi structures are pretty random, they should mostly be u32 or u64, with the occasional i32 - there's no reason for any f64 or f32 types, and certainly not both.  We should avoid usize completely.

This has a bunch of knock-on effects: we get to define JobID, Uid, and Pid as u64 which propagates through everywhere too.

It's too much to do this to fold it into #252 (at this stage of that patch stack, at least) so do it later.

lars-t-hansen:
This cleans up the types used in the GPU API.  That lets us get rid of a lot of casts elsewhere (though a few new ones need to be inserted, and there will be a followup bug for this one, cleaning up types used in `ps` especially).  Some types were factored out into small `types` module to create a shared bottom layer for all the subcommands.

I also ran cargo fmt and cargo clippy to make sure things were as good as they could be, this meant inserting some clippy overrides here and there.

This should be squashed when landing, as there's some underlying noise in the history that's a result of my bungling my fork.


lars-t-hansen:
I feared as much, the C-string literal syntax is not supported by our Rust version.

lars-t-hansen:
> I feared as much, the C-string literal syntax is not supported by our Rust version.

Requires Rust 1.77, I'll back out that change.

PR #280, opened 2025-04-25
Fix #147 - Daemon mode, built-in Kafka exfiltration - take 2

### LINKED ISSUE #147: Sonar "daemon" mode with Kafka exfiltration

lars-t-hansen:
This is a cleaned-up stack for https://github.com/NordicHPC/sonar/pull/255.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ With the branches come some additional rules for how to move patches around:
 
 At the time of writing we require:
 - 2021 edition of Rust
-- Rust 1.65.0 (can be found with `cargo msrv find`)
+- Rust 1.67.0 (can be found with `cargo msrv find`)
 
 Policy for changing the minimum Rust version:
 - Open a GitHub issue and motivate the change
--- a/doc/HOWTO-DAEMON.md
+++ b/doc/HOWTO-DAEMON.md
@@ -81,6 +81,8 @@ started, in addition to according to the cadence.
 cadence = <duration value>
 window = <duration value>                       # default 2*cadence
 uncompleted = <bool>                            # default false
+delta-coding = <bool>                           # default true
+dump = <filename>                               # default none
 ```
 
 The `window` is the sacct time window used for looking for data.
@@ -89,6 +91,21 @@ The `uncompleted` option, if true, triggers the inclusion of data about pending 
 This will result in multiple transmissions of data for the same `(job_id,job_step)`, one at each
 sample point.  If a job stays in, say, the PENDING state for several sampling windows then multiple
 transmissions for the job in the PENDING state will be seen.
+
+The `delta-coding` option, if true, triggers optimization of data transmission: redundant data are
+omitted in subsequent transmissions, and if there are no pertient data changes - for example, if a
+new `PENDING` record has the same contents as one already sent because no settings have changed -
+then the second record will not be sent at all.  What is deemed pertient or redundant is up for
+discussion for both `PENDING` and `RUNNING` jobs.  In any case, the recipient must be prepared to
+reconstruct the complete data stream from the initial record for the `(job_id,job_step)` and any
+subsequent deltas, applied in order.  If `uncompleted` is true, the `delta-coding` option can
+greatly reduce transmitted data volume.  If `uncompleted` is false, it can still avoid redundantly
+sending information about completed jobs.
+
+Setting `dump` to a file name will cause JSON output - without delta encoding - to be appended to
+the given file, while also being sent in the normal way.  If appending fails, a diagnostic is
+printed if `[debug]:verbose` is true.  (This is mostly useful for debugging the delta coding itself
+and to understand how to reconstruct the coded output.)
 
 ### `[cluster]` section
 
--- a/doc/HOWTO-DAEMON.md
+++ b/doc/HOWTO-DAEMON.md
@@ -6,6 +6,7 @@ command line parameter is the name of a config file.
 The daemon is a multi-threaded system that performs system sampling, communicates with the sink, and
 handles signals and lock files.
 
+If a sink is configured (currently only Kafka) then that is used for all data and control messages.
 If no other data sink is specified, the daemon prints output on stdout and reads control messages
 from stdin, see later sections.
 
@@ -24,10 +25,11 @@ divide an hour evenly and < 60, and hour values must divide a day evenly or be a
 of 24.  (Some sensible cadences such as 90m aka 1h30m are not currently expressible.)
 
 The config file has `[global]` and `[debug]` sections that control general operation; an optional
-section for the transport type chosen, currently none; and a section each for the sonar operations,
-controlling their cadence and operation in the same way as normal command line switches.  For the
-Sonar operations, the cadence setting is required for the operation to be run, the command will be
-run at a time that is zero mod the cadence.
+section for the transport type chosen, currently `[kafka]` (otherwise terminal I/O is used for
+transport); and a section each for the sonar operations, controlling their cadence and operation in
+the same way as normal command line switches.  For the Sonar operations, the cadence setting is
+required for the operation to be run, the command will be run at a time that is zero mod the
+cadence.
 
 ### `[global]` section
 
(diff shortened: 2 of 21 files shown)


README.md
```diff
@@ -55,7 +55,14 @@ It's sensible to run `sonar ps` every 5 minutes on every compute node.
...
 Take a snapshot of the currently running processes
 
-Usage: sonar ps [OPTIONS]
+Usage: sonar <COMMAND>
+
+Commands:
+  ps       Take a snapshot of the currently running processes
+  sysinfo  Extract system information
+  slurm    Extract slurm information
+  daemon   TODO!
```
lars-t-hansen, line 64: Flesh out here, and maybe add section below (TBD).

doc/HOWTO-DAEMON.md
```diff
@@ -0,0 +1,159 @@
...
+values are stripped.
+
+Boolean values are `true` or `false`.  Duration values express a time value using the syntax `__h`,
+`__m`, or `__s`, denoting hour, minute, or second values (uppercase HMS also allowed); values must
+be nonzero. For cadences, second values must divide a minute evently and be < 60, minute values must
+divide an hour evenly and < 60, and hour values must divide a day evenly or be a positive multiple
+of 24.  (Some sensible cadences such as 90m aka 1h30m are not currently expressible.)
+
+The config file has `[global]` and `[debug]` sections that control general operation; a section for
+the transport type chosen, currently only `[kafka]`; and a section each for the sonar operations,
```
lars-t-hansen, line 27: Oops, "kafka" here needs to be taken out, something else in its place; then "kafka" added later with that patch.

doc/HOWTO-DAEMON.md
```diff
@@ -0,0 +1,159 @@
...
+
+### `[global]` section
+
+```
+cluster = <canonical cluster name>
+role = node | master
+lockdir = <string>                              # default none
+```
+
+The cluster name is required, eg fox.educloud.no.
```
lars-t-hansen, line 40: Quote the name.

doc/HOWTO-DAEMON.md
```diff
@@ -0,0 +1,159 @@
...
+
+```
+cadence = <duration value>
+window = <duration value>                       # default 2*cadence
+uncompleted = <bool>                            # default false
+delta-coding = <bool>                           # default true
+dump = <filename>
+```
+
+The window is the sacct time window used for looking for data.
```
lars-t-hansen, line 99: Quote "window".

doc/HOWTO-DAEMON.md
lars-t-hansen, line 40: Quote "cluster", "role", "lockdir".

doc/HOWTO-DAEMON.md
```diff
@@ -0,0 +1,159 @@
...
+omitted in subsequent transmissions, and if there are no pertient data changes - for example, if a
+new `PENDING` record has the same contents as one already sent because no settings have changed -
+then the second record will not be sent at all.  What is deemed pertient or redundant is up for
+discussion for both `PENDING` and `RUNNING` jobs.  In any case, the recipient must be prepared to
+reconstruct the complete data stream from the initial record for the `(job_id,job_step)` and any
+subsequent deltas, applied in order.  If `uncompleted` is true, the `delta-coding` option can
+greatly reduce transmitted data volume.  If `uncompleted` is false, it can still avoid redundantly
+sending information about completed jobs.
+
+Setting `dump` to a file name will cause JSON output to be appended to the given file, while also
```
lars-t-hansen, line 116: This is also for delta-coding, along with the code implementing it, and should be (re)moved.

doc/HOWTO-DAEMON.md
```diff
@@ -0,0 +1,159 @@
...
+```
+
+### `[debug]` section
+
+```
+dump = bool                                     # default false
+verbose = bool                                  # default false
+```
+
+Setting `dump` to true will cause the daemon to dump all data it is sending on stdout.
```
lars-t-hansen, line 133: Not obvious that a dump to stdout is meaningful when stdio is the only i/o channel.  We could dump to stderr.  We could (re)move the function, which is really for the benefit of Kafka, and add it back for Kafka, but then in the Kafka section?

If we remove dump here and add it back elsewhere, possibly `verbose` should move to the global section.

doc/HOWTO-DAEMON.md
```diff
@@ -0,0 +1,159 @@
...
+is as configured in the `[global]` section and `<data-type>` is `sample`, `sysinfo`, `job`, `cluster`.
+
+The key sent with a message is currently the name of the originating node, including when that node
+is a master node.
+
+The values sent with these messages are opaque.  They may be a JSON object (always new-format JSON,
+see [NEW-FORMAT.md](NEW-FORMAT.md)), compressed text, and/or encrypted in some manner.  Currently
+there is no way of requesting anything other than JSON; if there is compression it is applied
+transparently, and if there is encryption it is by TLS, which must be configured explicitly in the
+config file.
```
lars-t-hansen, line 152: Reword the encryption thing since all of that was removed and will be something that comes after Kafka.

doc/HOWTO-DAEMON.md
```diff
@@ -0,0 +1,159 @@
...
+
+When using the stdout sink, the data messages are a JSON object with "topic", "key", and "value"
```
lars-t-hansen, line 154: Grammar.

doc/HOWTO-DAEMON.md
```diff
@@ -0,0 +1,159 @@
...
+is `node` or `master`.  These will have key and value as follows (very much TBD):
+
+```
+  Key     Value      Meaning
+  ------- ---------- -------------------------------------------
+  exit    (none)     Terminate sonar immediately
+  dump    <boolean>  Enable or disable data dump (for debugging)
+```
+
+TODO: The whole thing with control messages is a little dicey.  We would like to have mechanisms
```
lars-t-hansen, line 170: This is not a TODO but something else.

src/daemon.rs
```diff
@@ -0,0 +1,891 @@
+#![allow(clippy::comparison_to_empty)]
+#![allow(clippy::len_zero)]
+
+// TODO in this file, all marked:
+//
+// Medium pri
+//  - implementation of deluge filtering logic
```
lars-t-hansen, line 7: Remove this line.

src/daemon.rs
```diff
@@ -0,0 +1,891 @@
...
+            Dur::Minutes(n) => n,
+            Dur::Seconds(n) => (n + 59) / 60,
+        }
+    }
+}
+
+// The daemon_mode() should return under these circumstances *only*:
+//
+// - it fails to parse the ini file
+// - it fails to read the api token or password files
```
lars-t-hansen, line 146: "... any specified files" is probably more appropriate, also merge with next line

src/daemon.rs
```diff
@@ -0,0 +1,891 @@
...
+// - it fails to read the tls files
+// - it fails to acquire the lock file
+// - it fails to setup an interrupt handler
+// - an exit control message is received from the broker
+// - a signal is received from the OS or user that signifies an exit condition
+//
+// There are some problems that are not handled well.  The credentials for tls are file names that
```
lars-t-hansen, line 153: This paragraph belongs elsewhere.

src/daemon.rs
```diff
@@ -0,0 +1,891 @@
...
+        thread::spawn(move || {
+            repeated_event(json_tags::DATA_TAG_CLUSTER, sender, Operation::Cluster, c);
+        });
+    }
+
+    // At the moment, with the ini having no sections for data sinks, the only possible data sink is
+    // the stdio sink.
+    let mut data_sink: Box<dyn DataSink> = Box::new(StdioSink::new());
+
+    // This can fail for a number of reasons, notably cert/key/passwd files may not be found or they
```
lars-t-hansen, line 244: Stale comment, should finesse.

src/datasink.rs
```diff
@@ -0,0 +1,85 @@
+use crate::daemon;
+
+use std::io;
+use std::sync::mpsc;
+use std::thread;
+
+// The DataSink abstracts the details and hides the specific data sink we use.  The data sink
+// always runs on a separate thread and receives outgoing traffic on one channel and posts any
+// incoming messages on another.  The sink may batch outgoing messages, and its network connection -
```
lars-t-hansen, line 9: The comment about channels is false.

src/testdata/daemon-config.txt
```diff
@@ -0,0 +1,18 @@
+# This is a valid test configuration.  Some of the spacing is wonky in order to test
```
lars-t-hansen, line 1: "daemon-config.txt" is a bad name for this.  "daemon-stdio-config.txt" might be better.  Plus a comment to that effect.

tests/daemon-local.cfg
```diff
@@ -0,0 +1,34 @@
+# Config file for testing from a workstation (node) against a local kafka broker.
```
lars-t-hansen, line 1: "daemon-local.cfg" is a bad name for this, and where do we use it that it needs to live in this directory?

doc/HOWTO-DAEMON.md
lars-t-hansen: I'm moving dump to the kafka code and keeping verbose here.

lars-t-hansen:
The rust-kafka library looks more and more like abandonware.  I have some work with rdkafka on w-271-rdkafka on my fork, it does not look terrible and is (for now) the better way forward, even though rdkafka is also not being updated frequently.

lars-t-hansen:
Obsolete, use #289 instead.

PR #281, opened 2025-04-30
Fix # 267 - replace lazy_static by once_cell

lars-t-hansen:


lars-t-hansen:
Hm, I have really borked my fork, time to get a new one...

PR #285, opened 2025-05-06
Fix #284 - optional topic prefix

### LINKED ISSUE #284: Add "prefix" setting for the daemon
For testing, it should be possible to prefix (say) the topic name with something distinguished.  So the kafka section (probably, see below) of the config should have a "prefix" setting (maybe) that is prepended to the topic: with prefix=test, the topic becomes test.fox.educloud.no.sample.

(Prefix, postfix, infix - discuss.)  @2maz 

It's possible this is in the "global" section and not a kafka things since even the stdio data sink uses topics currently.  It's possible the name is "topic-prefix" and not just "prefix".

lars-t-hansen:
This does several things:

- add an optional topic prefix, see #284 for design notes, and document it
- makes the stdio sink require a topic on control messages read from stdin
- make the stdio sink a little more prominent in documentation
- add a regression test case for the daemon code using the stdio sink, sadly this was forgotten previously

### DIFF
--- a/doc/HOWTO-DAEMON.md
+++ b/doc/HOWTO-DAEMON.md
@@ -6,6 +6,9 @@ command line parameter is the name of a config file.
 The daemon is a multi-threaded system that performs system sampling, communicates with the sink, and
 handles signals and lock files.
 
+If no other data sink is specified, the daemon prints output on stdout and reads control messages
+from stdin, see later sections.
+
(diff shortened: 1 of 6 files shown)


lars-t-hansen:
@bast, optional review.

bast:
Thank you!

Issue #271, opened 2025-04-07
rdkafka support (producer)


lars-t-hansen:
rdkafka is the better Kafka library but not currently the one I'm working with for the daemon PoC.  I had some partial code to implement it, including some Cargo.toml adjusments, which I include here.

[rdkafka.diff.txt](https://github.com/user-attachments/files/19637226/rdkafka.diff.txt)

lars-t-hansen:
On the branch w-271-rdkafka on my fork.  This is working very well so far - SSL and SASL support were added without a hitch.

bast:
Cool!

lars-t-hansen:
I'm making this a producer-only issue.  The motivation for a consumer side (for control messages) is pretty iffy at this time, I'll spin off a bug for that to track it.

PR #289, opened 2025-05-12
#271 - Kafka exfiltration via rdkafka

lars-t-hansen:
rdkafka producer code with ssl and sasl support, and demo code for ingesting data from kafka into a directory tree.

What remains here:

- [x] testing
- [x] review

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,95 @@
 version = 3
 
 [[package]]
+name = "autocfg"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
+
+[[package]]
+name = "cc"
+version = "1.2.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8691782945451c1c383942c4874dbe63814f61cb57ef773cda2972682b7bb3c0"
+dependencies = [
+ "shlex",
+]
+
+[[package]]
+name = "cmake"
+version = "0.1.54"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e7caa3f9de89ddbe2c607f4101924c5abec803763ae9534e4f4d7d8f84aa81f0"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "cty"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b365fabc795046672053e29c954733ec3b05e4be654ab130fe8f1f94d7051f35"
+
+[[package]]
+name = "equivalent"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
+
+[[package]]
+name = "futures-channel"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2dff15bf788c671c1934e366d07e30c1814a8ef514e1af724a602e8a2fbe1b10"
+dependencies = [
+ "futures-core",
+]
+
+[[package]]
+name = "futures-core"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05f29059c0c2090612e8d742178b0580d2dc940c837851ad723096f87af6663e"
+
+[[package]]
+name = "futures-task"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f90f7dce0722e95104fcb095585910c0977252f286e354b5e3bd38902cd99988"
+
+[[package]]
+name = "futures-util"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9fa08315bb612088cc391249efdc3bc77536f16c91f6cf495e6fbe85b20a4a81"
+dependencies = [
+ "futures-core",
+ "futures-task",
+ "pin-project-lite",
+ "pin-utils",
+]
+
+[[package]]
+name = "hashbrown"
+version = "0.15.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "84b26c544d002229e640969970a2e74021aadf6e2f96372b9c58eff97de08eb3"
+
+[[package]]
+name = "indexmap"
+version = "2.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cea70ddb795996207ad57735b50c5982d8844f38ba9ee5f1aedcfb708a2aa11e"
+dependencies = [
+ "equivalent",
+ "hashbrown",
+]
+
+[[package]]
+name = "itoa"
+version = "1.0.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4a5f13b858c8d314ee3e8f639011f7ccefe71f97f96e50151fb991f267928e2c"
 
 [[package]]
 name = "libc"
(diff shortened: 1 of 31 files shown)


lars-t-hansen:
Apparently we need to bump MSRV to deal with sending enums on channels, the Sync trait is not implemented (and it should be derived automatically).

lars-t-hansen:
1.72.1 is the new new thing.

src/daemon.rs
```diff
@@ -695,6 +772,11 @@ fn parse_config(config_file: &str) -> Result<Ini, String> {
         return Err("Missing global.role setting".to_string());
     }
 
+    #[cfg(feature = "kafka")]
```
lars-t-hansen, line 775: TODO: If have_kafka and sasl-password is not empty then ca-file must not be empty.

Cargo.lock
```diff
@@ -2,31 +2,290 @@
...
+]
+
+[[package]]
+name = "rdkafka"
+version = "0.37.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14b52c81ac3cac39c9639b95c20452076e74b8d9a71bc6fc4d83407af2ea6fff"
+dependencies = [
+ "futures-channel",
+ "futures-util",
```
lars-t-hansen, line 205: futures-util and futures-channel really should not be a dependency for baseconsumer and baseproducer but I don't yet know how to get rid of these.

doc/HOWTO-DAEMON.md
```diff
@@ -52,6 +53,24 @@ If there is a `topic-prefix` then it is prefixed to each data packet's topic.  A
...
+sending-window = <duration value>               # default 5m
+ca-file = <filename>                            # default none
+sasl-password = <string>                        # default none
+```
+
+The `broker-address` is required and names the address of the broker.  For Kafka it's usually
+host:port, eg `localhost:9092` for a local broker on the standard port.
+
+All available data are sent to the data sink at some random time within the `sending-window`, which
+starts at the point when there are no data available to send.
```
lars-t-hansen, line 69: Another History of the siege of Lisbon: remove the "no".

doc/HOWTO-KAFKA.md
```diff
@@ -0,0 +1,226 @@
...
+Test notes with standard Kafka server, see https://kafka.apache.org/quickstart.
+
+#### Setup
+
+You're going to be running several shells, let's call them Zookeeper, Server, Consumer, Work, and Sonar.
+
+The working directory for the following is the root directory of the unpacked Kafka distribution, eg
+`kafka_2.13-3.9.0/`.
+
+NOTE!  Currently Sonar does not work with Kafka 4.0.0, there's a protocol error, not debugged.  Use
```
lars-t-hansen, line 45: We might check if this is still true - that was a problem with kafka-rust, but librdkafka is probably better.  Either way, we might want to reword this.

doc/HOWTO-KAFKA.md
```diff
@@ -0,0 +1,226 @@
...
+
+```
+  bin/kafka-server-start.sh config/server.properties
+```
+
+In the Work shell, topics need to be added with `kafka-topics.sh` a la this, if you haven't done it
+before (or if you did it, but did not shut down the broker properly):
+
+```
+  bin/kafka-topics.sh --create --topic test-cluster.hpc.uio.no.sample --bootstrap-server localhost:9092
```
lars-t-hansen, line 66: move the bootstrap server argument to the left and break lines.

doc/HOWTO-KAFKA.md
```diff
@@ -0,0 +1,226 @@
...
+  test-cluster.hpc.uio.no.control.master
+```
+
+#### Running sonar and examining the data
+
+Then from the Sonar root directory, after building it, run Sonar in daemon mode with a suitable
+config file in the Sonar shell:
+
+```
+  target/debug/sonar daemon <config-file>
```
lars-t-hansen: A proper filename comes in a subsequent patch, so OK to leave this alone here.

doc/HOWTO-KAFKA.md
```diff
@@ -0,0 +1,226 @@
...
+```
+
+Sonar will run continuously and start pumping data to Kafka.
+
+For easy testing, run this in the Consumer shell to listen for `sysinfo` messages and echo them:
+
+```
+  bin/kafka-console-consumer.sh --topic 'test-cluster.hpc.uio.no.sysinfo' --bootstrap-server localhost:XXXX
```
lars-t-hansen, line 110: As above, and also below, move the bootstrap-server argument and break the lines.

doc/HOWTO-KAFKA.md
```diff
@@ -0,0 +1,226 @@
...
+We will need a CA certificate (to be used by both Kafka and Sonar) and a key store containing the
+server's public and private keys (to be used by Kafka).  NOTE, the server keys will be tied to a
+particular server name.
+
+For testing, we will generate our own CA and key materials.  In util/ssl, there is a Makefile that
+will generate the necessary files: `sonar-ca.crt` is the CA certificate, and
+`sonar-kafka-keystore.pem` is the key store.  Just run `make all` to make the files for the local
+hostname.
+
+Having generated those, update Kafka's `server.properties` as follows, this is almost in the form of
```
lars-t-hansen, line 160: Clean this up.  Move the diffs from the last patch to this patch, and rewrite the prose here to reflect the availability of those files.  Keep the note that the paths may need to be updated.

doc/HOWTO-KAFKA.md
```diff
@@ -0,0 +1,226 @@
...
+#### Authentication
+
+We will use SASL PLAIN authentication over SSL for now.  In this scheme, the server's config file
+contains information about security principals and their passwords.  We will use the cluster name as
+the principal name (this will come into play later when we implement authorization).
+
+We can use the same key materials we generated above for SSL, but there are additions to the config
+file and the daemon's .ini file.
+
+Changes to `server.properties` relative to the above:
```
lars-t-hansen, line 210: Same as above.

doc/HOWTO-KAFKA.md
```diff
@@ -0,0 +1,226 @@
...
++
++sasl.enabled.mechanisms=PLAIN
++listener.name.client.plain.sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required \
+   user_test-cluster.hpc.uio.no="test-password";
+```
+
+Note that in that last phrase, `client` is the same as `CLIENT` higher up but must here be in lower
+case for reasons unknown.
+
+The .ini file gets an addition in the `[kafka]` section: the new `sasl-password` property must hold
```
lars-t-hansen, line 230: Same as above.

doc/HOWTO-KAFKA.md
```diff
@@ -81,7 +81,13 @@ Then from the Sonar root directory, after building it, run Sonar in daemon mode
 config file in the Sonar shell:
 
 ```
-  target/debug/sonar daemon <config-file>
+  target/debug/sonar daemon tests/kafka/sonar-slurm-node.ini
```
lars-t-hansen, line 86: These changes can and should be moved to the previous patch, since the ini files are available there.

doc/HOWTO-KAFKA.md
```diff
@@ -4,6 +4,8 @@ In the "daemon mode", Sonar stays memory-resident and pushes data to a network s
 sinks is a Kafka broker.  See HOWTO-DAEMON.md for general information about this mode and the
 options available for configuring the Kafka producer in Sonar.
 
+See `../tests/kafka/` for examples of config files for various node and master types.
```
lars-t-hansen, line 7: This can be moved to the first patch because the ini files are being moved there.

doc/HOWTO-KAFKA.md
```diff
@@ -92,7 +94,17 @@ And/or on a single node with access to slurm (eg a login node):
 
 Sonar will run continuously and start pumping data to Kafka.
 
-For easy testing, run this in the Consumer shell to listen for `sysinfo` messages and echo them:
+In the Consumer shell, go to `util/ingest-kafka` and build `ingest-kafka` if you haven't already.  Run
```
lars-t-hansen, line 97: This comment should be moved to the ingest-kafka patch.

tests/kafka/server.properties-with-ssl-sasl.diff
```diff
@@ -0,0 +1,34 @@
+*** server.properties.orig	2025-05-12 14:13:30.855032598 +0200
```
lars-t-hansen, line 1: The diffs should be moved to the first patch, because that's where we explain how to set these properties.

lars-t-hansen:
@bast, PTAL.  I'm still working on the manual Kafka test case (last commit - it is not stable though it works some of the time) but the prior two patches should be done.  In truth only the first patch is important.

lars-t-hansen:
Added some more time lapse to the test case, seems OK - no promises, but it doesn't run in automation, so it's probably OK.

bast:
Thank you! Looking ...

bast (approved):
I read the documentation and it is clear. Otherwise probably not super easy for me to test it in addition but I don't think it's needed.

Thank you for the work!

lars-t-hansen:
Thanks.  I still chafe at the number of dependencies and contributors that are pulled in by this patch but only kafka depends on them, at least.

PR #292, opened 2025-05-14
Fix #274 - check Rust and Go code formatting

### LINKED ISSUE #274: Test that formatting is good on every PR
No PR should land without proper formatting, and we should be able to integrate this as a guard on the PR, as we do now with functionality tests.

For the Go code, I think it's simply running `test $(gofmt -l util) == ""`

For the Rust code, it's TBD.

We can discuss whether this is something that could even be integrated into the test suite so that `make test` checks it, in local testing, I might prefer that (and test.yml could just run `make test`, but TBD).  Maybe have a `make integration` so that `make test` just runs functionality tests.

lars-t-hansen:
The checks for proper format are both fast (given how tiny the repo is) so I've broken them out as separate workflows.  Also included are some reformats, of course, since otherwise this PR would not pass the tests...  The new tests have to be run manually from the tests directory, they are not integrated into run_tests.sh, see the bug for discussion.

### DIFF
--- a/.github/workflows/test-go-format.yml
+++ b/.github/workflows/test-go-format.yml
@@ -1,0 +1,24 @@
+name: Test Go code formatting
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+
+    - name: Display Go version
+      run: go version
+
+    - name: Format code
+      run: |
+        cd tests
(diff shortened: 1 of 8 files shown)


lars-t-hansen:
As desired, when code is not formatted correctly we get an error in the workflow:
```
FORMATTING FAILURE!
The following files are not properly formatted (cargo fmt):
/home/runner/work/sonar/sonar/src/datasink.rs
/home/runner/work/sonar/sonar/src/slurmjobs.rs
/home/runner/work/sonar/sonar/src/sysinfo.rs
Error: Process completed with exit code 1.
```

bast:
Thank you!

PR #299, opened 2025-05-16
test: add Docker configuration for broker and ui with ssl_sasl config

2maz:
Finalized the example docker setup.


doc/HOWTO-KAFKA.md
```diff
@@ -37,9 +37,14 @@ Test notes with standard Kafka server, see https://kafka.apache.org/quickstart.
 
 #### Setup
 
+Download a [Kafka Binary Release](https://dlcdn.apache.org/kafka/3.9.0/kafka_2.13-3.9.0.tgz) and unpack it:
```
lars-t-hansen, line 40: For transparency and some resilience, could we instead link to https://kafka.apache.org/downloads and state in text that one should for now download the 3.9.0 release as that's the only that is tested and supported?

util/ssl/kafka-jaas.conf
```diff
@@ -0,0 +1,13 @@
+KafkaServer {
```
lars-t-hansen, line 1: Strikes me that a more natural location for this file is with the compose file that uses it?  See the overall review comment for something related.

lars-t-hansen (changes requested):
This is great, and I look forward to testing it myself, but some general suggestions:

- could you add some appropriate prose to README.md in this directory (similar to the prose for the existing test case) so that people who come here know what to do with these files and how to run the tests properly?
- more generally, since the new test is independent of the existing test (I think?) we should probably create subdirectories here for each test with a README in each.  That way, any directory structure needed by docker to hold the jaas file could also be in that subdirectory, I think.  After all the jaas file is part of the kafka config, not strictly related to the crypto stuff, and does not belong in util/ssl I think.  If any files are needed in the docker's file system that are now in the ssl directory they can be copied maybe by the test script?

lars-t-hansen (approved):
Alright.  The

PR #314, opened 2025-05-28
Fix #312 - build nvidia-on-arm64

### LINKED ISSUE #312: gpuapi build scripts for aarch64-nvidia on ex3
On ex3 we actually have aarch64 nodes with nvidia gpus and we should create a new build script for the gpuapi lib to build this library on those nodes (eg gh001).

lars-t-hansen:
Build script for the nvidia gpuapi on aarch64 on ex3 (only on gh001 at the moment), and the library thus built.


gpuapi/build-ex3-nvidia-aarch64.bash
2maz, line 1: Suggestion:
```suggestion
#!/usr/bin/env bash
#
# See Makefile for information.
#

set -e
module purge

GPU_TYPE=''
ARCH=$(uname -m)
if `command -v nvidia-smi`; then
    if [ "$(nvidia-smi -L 2> /dev/null | wc -l)" -gt 0 ]; then
        GPU_TYPE="nvidia"
        module load cuda12.8
    fi
fi

if `command -v rocm-smi`; then
    # ignore header line
    if [ "$(rocm-smi --showuniqueid --csv | awk NF | wc -l)" -gt 1 ]; then
        GPU_TYPE="amd"
    fi
fi

if [ "$GPU_TYPE" == '' ]; then
    echo "Neither Nvidia nor AMD GPU found"
    exit 1
fi

module list
make libsonar-$GPU_TYPE.a
mkdir -p $ARCH
mv libsonar-$GPU_TYPE.a $ARCH

echo ""
echo "Available for architecture $ARCH:"
ls $ARCH

lars-t-hansen: I don't think that's an appropriate fix for a script whose name contains the substring "nvidia", for one thing.  More generally, the current system is that each of these scripts builds for one GPU type on one CPU architecture type - since each build has to happen on an appropriate host.  We could generalize that, but then I'd like that to be a separate work item, because it'll be a bit of work to design and test.  Right now I would prefer to make incremental progress using the existing setup.

2maz: For the eX3 script I would not want to list (and maintain) all hostnames in here, thus I suggest to check GPU availability via checking for nvidia-smi. On eX3 we have to cover most combinations of intel|arm - nvidia | amd.
So we can have a script per combination or generalize in one script - your choice - suggested functions for that are in the comment.

There seems to be no different architecture handling needed except for the library location. So I suggest to use ARCH=$(uname -m) for the folder part.

lars-t-hansen: The purpose of listing specific hosts is to ensure that the library is being compiled on a host that is known to be a good compilation host.  We could generalize the pattern but that code is not there for nothing.

My experience so far with the smi programs is that the options and outputs change from release to release, hence my comment about "a bit of work to design and test".

In light of this discussion, I'm going to land this patch as-is on 0.13 and main and then file a followup bug to generalize the build scripts with a reference to this PR.

PR #322, opened 2025-06-05
Fix #300 - add 'directory' output path to daemon mode

### LINKED ISSUE #300: Add data sink for directory tree
We want `sonar daemon` to replace the `sonard` utility in Jobanalyzer. We should add a data sink that dumps data into a directory tree that is compatible with jobanalyzer (and also with traditional sonar storage).

lars-t-hansen:


lars-t-hansen:
Patch looks OK and can land once the runners are done.

PR #323, opened 2025-06-12
Fix #249 - Record load average

### LINKED ISSUE #249: Load averages
(To be fleshed out)

Similar to #242, we want to exfiltrate the node load average as a proxy for the number of runnable jobs, to give an idea of imbalance between allocated resources and available work.

lars-t-hansen:


lars-t-hansen:
Needs a test case, I guess.

src/procfs.rs
```diff
@@ -252,6 +252,34 @@ pub fn get_node_information(system: &dyn systemapi::SystemAPI) -> Result<(u64, V
...
+    let entities = fields[3].split('/').collect::<Vec<&str>>();
+    if entities.len() != 2 {
+        return Err("Unexpected entities".to_string());
+    }
+    let runnable = entities[0]
+        .parse::<u64>()
+        .map_err(|_| "Bad entity".to_string())?;
+    let existing = entities[1]
+        .parse::<u64>()
+        .map_err(|_| "Bad entity".to_string())?;
```
lars-t-hansen, line 279: The error messages are not good - the last two are worse than the others, too.  Basically these will never happen, but if they do happen we want maximal information, so: better prose, and include the input line in the messages.

PR #334, opened 2025-06-24
Fix #326 - properly document rolled-up samples

### LINKED ISSUE #326: Document rolled-up samples
There are documentation snippets scattered across the sonar sources, the sonar format spec, and the jobanalyzer sources that together define exactly what the rolledup field means and what its invariants are.  TL;DR: it's too complicated.  But since this field is present in all historical and current data from Sonar, we should extract that documentation into the format spec.  The rolledup field probably still pays for itself on the big systems.

lars-t-hansen:


bast:
Thank you!

PR #350, opened 2025-08-29
Fix #286, #348, #349

### LINKED ISSUE #286: Switch from mpsc to crossbeam
Currently the daemon uses sync::mpsc for inter-thread communication, but that is a very limited interface - it has no selection facility, for one thing.  (My Go code that implements a similar daemon is significantly more elegant.)  The advice I find on the net is to move to [crossbeam](https://docs.rs/crossbeam/latest/crossbeam/index.html) instead, another v0.something third-party crate with at least one external dependency - hardly Rust's finest moment, even if many of the crossbeam contributors are part of the Rust core team.  Crossbeam has selection, clock tick channels, and timeout channels, all of which we want here.

### LINKED ISSUE #348: Daemon mode must handle SIGTERM
On most systems sonar will be managed in daemon mode by systemd, and systemd will send SIGTERM when a stop or restart command is issued.  The daemon mode must catch and attempt to respond to SIGTERM in a clean fashion within 90s.  Basically, when SIGTERM is received, sonar should stop producing data, and should instruct the data sink to flush if possible.  When the sink has flushed, sonar can exit.  If flushing fails, sonar will be killed by systemd after 90s, but this is OK - we would lose data anyway.

This is related to #301 but is important in its own right.

### LINKED ISSUE #349: Cleanup: Signal handling for `ps` should use signal_hook
For #348 I decided to resort to the signal_hook crate rather than digging myself further into the hole with homebrew signal handling logic, it becomes involved when data have to be communicated and a thread needs to block on signals arriving.  Since we now have a dependency on signal_hook we might as well use it for the signal handling in the `ps` subcommand, and get rid of our own signal handler code.  At this stage we pretty much have no other choice than accepting lots of dependencies anyway.

lars-t-hansen:
Fix #286 - mpsc -> crossbeam::channel; make Kafka interface use it
Fix #348 - signal handling for daemon via signal_hook crate
Fix #349 - handle signals in ps via signal_hook crate
Removes existing signal handling module
Test scaffolding
Test cases - many new and some cleaned up


src/datasink/kafka.rs
```diff
@@ -175,78 +187,96 @@ fn kafka_producer(
...
+        ) {
+            Ok(()) => {}
+            Err(m) => {
+                // An error here only means that the message could not be enqueued; sending errors
+                // are discovered in the ProducerContext.  So an error here is pretty much fatal for
+                // the message, hence we drop it.
+                let msg = format!("Message production error {m}");
+                let _ = control_and_errors.send(Operation::MessageDeliveryError(msg));
+                if !flushing {
+                    break 'sender_loop;
```
lars-t-hansen, line 267: Probably should add a comment here that breaking out of the loop is optional - it's not obvious what the benefit is, and in fact, if there are "enough" bad messages then the result will be that the backlog will become longer and longer.  The test case has some comments relating to this too.  It's possible we should just fix it now.

lars-t-hansen:
@bast, FYI in case you want to take a look.  I've given up on supply chain security - I'll never get anywhere that way, and with the previous rdafka dependency that ship sailed anyway.

PR #355, opened 2025-09-04
Add yaml spec generation

2maz:
The spec file is intended to be used for validation of the table spec implementation, e.g., as in slurm-monitor's sqlalchemy definitions

Example output
```
# AUTOMATICALLY GENERATED BY `process-doc -w -yaml`.  DO NOT EDIT.
# Edit `util/formats/newfmt/types.go`, then in `util/process-doc` run `make install`.

MetadataObject:
  fields:
    producer:
      type: 'NonemptyString'
      doc: 'The name of the component that generated the data (eg "sonar", "slurm-monitor")'
    version:
      type: 'NonemptyString'
      doc: 'The semver of the producer'
    format:
      type: 'uint64'
      doc: 'The data format version'
    attrs:
      type: '[]KVPair'
      doc: 'An array of generator-dependent attribute values'
    token:
      type: 'string'
      doc: 'EXPERIMENTAL / UNDERSPECIFIED.  An API token to be used with Envelope.Data.Attributes.Cluster, it proves that the producer of the datum was authorized to produce data for that cluster name.'
``

### CODE REVIEW

File: util/process-doc/process-doc.go [hunk 1]
```diff
@@ -44,6 +45,10 @@ func main() {
 	case *makeRust:
 		fmt.Print("// AUTOMATICALLY GENERATED BY `process-doc`.  DO NOT EDIT.\n")
 		fmt.Print("// Edit `util/formats/newfmt/types.go`, then in `util/process-doc` run `make install`.\n\n")
+
+	case *makeYaml:
+		fmt.Print("# AUTOMATICALLY GENERATED BY `process-doc -w -yaml`.  DO NOT EDIT.\n")
```
lars-t-hansen, line 50: It's a good idea to specify in this comment that the -yaml option was used, and it would be great if you could add the similar clarification to the two previous blocks while you're in here.

Probably you want to remove the -w in the text since there's no reason to assume one way or the other that it was used, and -w should not have any bearing on the output.

util/process-doc/process-doc.go
```diff
@@ -159,23 +176,33 @@ func handleTypeAlias(item *ast.GenDecl, td *ast.TypeSpec) {
...
-				ndoc = strings.TrimPrefix(doc, "//")
-			}
-			if !todo && strings.Index(ndoc, "TODO") >= 0 {
-				warn(ctx, "TODO in comment")
-				todo = true
-			}
-			fmt.Println(ndoc)
-		}
+		var docs = extractDocs(ctx, d, "\n")
+		fmt.Println(docs)
```
lars-t-hansen, line 205: Since docs is only used once you can just fold the call into the println argument list.

util/process-doc/process-doc.go
```diff
@@ -159,23 +176,33 @@ func handleTypeAlias(item *ast.GenDecl, td *ast.TypeSpec) {
...
 	} else {
-		var todo bool
-		for _, c := range d.List {
-			doc := c.Text
-			ndoc := strings.TrimPrefix(doc, "// ")
-			if ndoc == doc {
-				ndoc = strings.TrimPrefix(doc, "//")
-			}
-			if !todo && strings.Index(ndoc, "TODO") >= 0 {
-				warn(ctx, "TODO in comment")
```
lars-t-hansen, line 174: This warning was lost in translation.

lars-t-hansen (changes requested):

util/process-doc/process-doc.go
2maz: indeed, added - also added the ndoc output to show the actual todo

lars-t-hansen (approved):
LGTM.  Before landing, please run `make install` here so that all output files are in sync with the script, and add the generated files to the commit.  Also please squash changes when landing.

PR #359, opened 2025-09-15
Fix #207 - XPU support

### LINKED ISSUE #207: Add XPU support
I have access to an XPU system and I need XPU support for a NAIC use case - rough parity with nvidia and amd support.  (Sonar is growing a few features so that it can take the place of another monitoring component running on a partner system.)

lars-t-hansen:


2maz (approved):
I see get_card_uuid_from_info is only introduced for amd+nvidia, is xpu deviating on purpose?

PR #364, opened 2025-09-16
Allow for architecture specific features

2maz:


2maz:
> I need to think about this for a bit. It's generally a complication though of course it should work. What do we gain by it?

we gain building on aarch64 without errors, since otherwise I run into a linker error looking for a missing xpu library.
There are probably other ways to handle this via cargo features.

lars-t-hansen:
y, i realized this last night too.  the alternative at the moment is to build stub gpu libraries for habana and xpu on aarch64, the code is set up for that at present and all that's needed is some support down in gpuapi/.  it's what we do for amd on aarch64 at present (not having access to build hardware with that combination).

Probably going forward, for a long while,

- we won't be adding many more gpus, if any
- we won't be adding any more cpu architectures

so we could solve it either with a build fix or a stub fix, but it needs a proper solution.

Changing the Makefile is not the best fix (or at least not a complete fix) because most test cases don't build via the Makefile but use cargo build or cargo run, and the test cases for xpu and habana currently assume that the features are enabled by default.  So minimally they would need to enable those features.  But then they may need to be gated on x86_64.

lars-t-hansen:
I think, for the time being it's easier to fix this with the necessary stub libraries.  I have a patch already that makes linking work on aarch64 by adding the XPU library, and I'll be sure to do the same for Habana.

PR #375, opened 2025-09-19
Fix #363 - Update rdkafka, once_cell, Cargo.lock

### LINKED ISSUE #363: Compilation fails on RHEL10 - openssl incompatibility
This is probably an issue with changes to the OpenSSL header files in openssl-devel.  When compiling rdkafka, the C compiler fails:
```
  rdkafka_admin.c:5909:21: error: implicit declaration of function ‘RAND_priv_bytes’ [-Wimplicit-function-declaration]
   5909 |                 if (RAND_priv_bytes(random_salt, sizeof(random_salt)) == 1) {
        |                     ^~~~~~~~~~~~~~~
```
We can fix that by including openssl/rand.h in rdkafka_admin.c.

But then there are further failures in rdkafka_ssl.c where ENGINE_whatever definitions are not found.  These names don't seem to be defined any longer.  /usr/include/openssl/engine.h is empty.  This appears to be because the API is gone in RHEL10, https://github.com/dotnet/runtime/issues/104775, and the package that papered over this (openssl-devel-engine) is Fedora-only.

Gemini says "It is important to note that the OpenSSL Engine API has been deprecated in OpenSSL version 3.0."  Joy.  The horse's mouth: https://docs.openssl.org/3.6/man3/ENGINE_add/#synopsis.

Since rdkafka and librdkafka don't seem to be very actively maintained any longer, this is an existential problem for us - the rdkafka maintainers are likely not going to fix it, but we need it to be fixed . In the very short term we may be able to work around it by compiling on older systems.  But even in the short/medium term we should expect systems to move to RHEL10 or similar.

lars-t-hansen:


lars-t-hansen:
Naturally this upgrade requires the minimal rust version to change (to 1.76)...

- On the ML nodes, we already have to install Rust for the user since the most recent Rust is 1.70.
- On Fox, the newest Rust is 1.85.
- On Saga, it is 1.83.
- On Betzy, it is 1.78.
- On eX3 the latest rust is generally 1.85.
 
Since the current Rust version is 1.90, it feels OK to upgrade the MSRV.

Updating to 1.76 allows us to use div_ceil (introduced in 1.73).  Updating to 1.77 would also allow the use of "c" literals, maybe worth doing while we're in here?

PR #389, opened 2025-09-26
Fix #336 - test cases for sacct, sinfo parsing

### LINKED ISSUE #336: Test cases for sacct and sinfo parsing
The running of both sacct and sinfo has been virtualized properly but I don't think I have any test cases for the parsers; we should have that, at least to guard against regressions.

lars-t-hansen:


lars-t-hansen:
This is blocked on #388, will rebase to main once that lands and do the test runs.

PR #393, opened 2025-10-03
Fix #335 - test case for cpu_util + fixes

### LINKED ISSUE #335: Test case for cpu_util computation
The CPU utilization is computed by sampling CPU time (in ticks), waiting a short while, and then sampling again and from that we get an "instantaneous" reading of CPU usage.  The code is not complicated but there's no test case for it, and we should have one.

lars-t-hansen:
Going to ponder this further over the weekend, but I think this is roughly right.

### DIFF
--- a/doc/NEW-FORMAT.md
+++ b/doc/NEW-FORMAT.md
@@ -661,14 +661,14 @@ NOTE: Container software may not reveal all the process data we want. Docker, fo
 provides cpu_util but not cpu_avg or cpu_time, and a memory utilization figure from which
 resident_memory must be back-computed.
 
-NOTE: The fields cpu_time, cpu_avg, and cpu_util are different views on the same
-quantities and are used variously by Sonar and the slurm-monitor dashboard. The Jobanalyzer
-back-end computes its own cpu_util from a time series of cpu_time values and using the
-cpu_avg as the first value in the computed series. The slurm-monitor dashboard in contrast
-uses cpu_util directly, but as it will require some time to perform the sampling it slows down
-the monitoring process (a little) and make it more expensive (a little), and the result is less
-accurate (it's a sample, not an averaging over the entire interval). Possibly having either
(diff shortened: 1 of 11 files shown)


doc/NEW-FORMAT.md
```diff
@@ -328,11 +328,13 @@ Primary memory in kilobytes
 
 #### **`topo_svg`** string
 
-Base64-encoded text output (SVG source text) of `lstopo` or similar
+Base64-encoded text output (SVG source text) of `lstopo` or similar, using the base64
+alphabet from RFC4648.
```
lars-t-hansen, line 332: Probably land this and related fixes separately.

tests/Makefile
```diff
@@ -1,7 +1,7 @@
-CFLAGS=-Wall -std=c89
+CFLAGS=-Wall -std=c89 -O1
 .PHONY: all clean
 
-all: rollup rollup2 rollupchild rollupchild2
+all: rollup rollup2 rollupchild rollupchild2 pincpu
```
lars-t-hansen, line 4: Move pincpu to the front so that all the lists are the same.  Possibly, `all` is the wrong idea and pincpu does not belong in this list at all, since the others are needed for the rollup tests.

tests/ps-cpu-util.sh
```diff
@@ -0,0 +1,37 @@
...
+#
+# We want the cpu_util field from the process with cmd=pincpu (ideally where the enclosing job has
+# user=$LOGNAME).  Ignore the bit with $LOGNAME for now.
+#
+# The cpu_util is floating point, so we need to round it.
+
+output=$(cargo run -- ps --exclude-system-jobs --json --cluster my.cluster)
+util=$(jq '.data.attributes.jobs[].processes[]|select(.cmd=="pincpu").cpu_util|ceil' <<< $output)
+
+if (( util < 75 )); then
```
lars-t-hansen, line 32: Don't we want an upper bound too?

Issue #384, opened 2025-09-24
Test case timing issue on heavily loaded systems


lars-t-hansen:
See #382.  On heavily loaded systems, tests/lockfile.sh will fail because the timeouts are not long enough / synchronized properly.  This is not so easy to fix, necessarily.

PR #412, opened 2025-10-17
Fix #408 - default to new JSON format

### LINKED ISSUE #408: Default to new output format
Increasingly, new fields are not added to the "old" output format and some subcommands don't even have an old format.  This bites back from time to time when users are not aware of this.  At this stage we should switch to the new output format by default and allow the old to be selected by switch (see also #244 which removes the old CSV formats entirely).

Tentatively, I think we should backport this to 0.16 even though that is in rc, there's no sense in waiting.

In other words:

- the default is as for `--json` now: new JSON output format
- for runs from the command line we should allow there to be no --cluster option, and there should be a sensible default string ("no.cluster")
- the --json switch should do nothing
- the old format should be selectable with a switch --oldfmt

lars-t-hansen:


src/linux/system.rs
```diff
@@ -43,7 +43,7 @@ impl Builder {
     pub fn new() -> Builder {
         Builder {
             jm: None,
-            cluster: "".to_string(),
+            cluster: "no.cluster".to_string(),
```
lars-t-hansen: Maybe the empty string would have been better after all?

lars-t-hansen:
There was an intermittent failure on one run, looks like a timing problem.  See #384.

PR #416, opened 2025-10-18
Fix #413 - clean up gpu tests

### LINKED ISSUE #413: GPU tests still depend on old output format
When the tests for #408 landed (also on 0.16) the GPU tests were not properly updated to deal with the new default format because they are not run except on GPU nodes and who has the time to login everywhere and run it?  (Running it on one node at least would have been sensible.)

This need be nothing more than adding the --csv switch, but really a number of those tests would be better if they used the JSON format, so it may be worth actually cleaning that up.

lars-t-hansen:


tests/amd-gpu.sh
```diff
@@ -32,11 +34,10 @@ fi
...
 # TODO: This will be cleaner once we have json output.
 
-output=$(cargo run -- ps --load --exclude-system-jobs)
-infos=$(grep -E 'gpuinfo=.*fan%=.*tempc=.*' <<< $output)
-lines=$(wc -l <<< $infos)
+cargo run -- ps --load --exclude-system-jobs --csv > $output
+lines=$(grep -E 'gpuinfo=.*fan%=.*tempc=.*' < $output | wc -l)
 if (( lines != 1 )); then
-    fail "Number of matching output lines should be 1"
+    fail "Number of matching output lines should be 1, go $lines"
```
lars-t-hansen: "go" => "got"

tests/shared-gpu-smoketest
```diff
@@ -0,0 +1,42 @@
...
+# To be sourced from the GPU tests.
+# Free variables:
+#  output - name of temp file
+
+set -o pipefail
+
+# Test that sysinfo finds the cards.  This is also sufficient to test that the GPU SMI library has
+# been found and is loaded.
+
+# nvidia is enabled by default
```
lars-t-hansen: Stray comment.

PR #422, opened 2025-10-22
Fix #409 - slurm resource parsing

### LINKED ISSUE #409: Resource allocation not shown for PENDING slurm jobs
From @2maz I hear that sacct will not print the resource allocation request for pending jobs, but we do want to see that.  The word is that we have to run scontrol to get that information.  Everything else is TBD.

Will need to see how broad this change is before deciding whether to backport to 0.16 or delay until 0.17.

lars-t-hansen:
The main changes are:

- run scontrol in addition to sacct for `sonar slurm` to get request data for pending/running jobs
- if we have ReqTRES for a job from scontrol, fill in the gres_detail field with this
- otherwise if we have AllocTRES for a job from sacct, fill in the gres_detail field with this
- remove the minimum_cpus_per_node field because we weren't using it and had no semantics for it

The rest is plumbing and test cases.


lars-t-hansen:
Looking at the output, a small number of CANCELLED jobs have no gres_detail value.  (Most CANCELLED jobs do.)  These are probably jobs that were cancelled while still pending, because in that case, they disappear from the scontrol output before they are given an allocation.  The AllocTRES from sacct will be empty, and we have no fallback from scontrol.  I'm not going to worry about this now.

2maz:
@lars-t-hansen  lgtm - FYI: as soon as the job is completed or cancelled the gres_detail might vanish (from scontrol) - at least that was what I observed when using slurmrestd

PR #436, opened 2025-10-31
Fix #433 - actually split output into messages

### LINKED ISSUE #433: Implement batching of slurm jobs data
In an effort to alleviate #276, implement simple batching of slurm jobs data.  The batch size is by default unlimited but is to be settable by `--batch-size` on the command line or by `jobs.batch-size' in the daemon config file.

lars-t-hansen:
Actually not too horrifying - there's some plumbing so that the run() methods all return vectors of vectors of output, the subvectors representing individual messages.  Then some plumbing to fit that into existing code most places, only the slurmjobs module needs to use it.  Then code in the daemon to process each message, not just the one.  Then a debug setting for one-shot runs so that we can write a sensible test.  And then the test case.

### DIFF
--- a/doc/HOWTO-DAEMON.md
+++ b/doc/HOWTO-DAEMON.md
@@ -177,6 +177,8 @@ cadence = <duration value>
 ```
 verbose = bool                                  # default false
 time-limit = <duration value>                   # default none
+oneshot = bool                                  # default false
+output-delay = <duration value>                 # default none
 ```
 
 Setting `verbose` to true will cause the daemon to print somewhat informative messages about what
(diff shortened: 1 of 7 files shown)


lars-t-hansen:
This is currently running on los.fox against the latest sonalyze; so far so good, but need to let it go for a few days to check that batching works and that there are no sending errors.

lars-t-hansen:
```
$ jq .data.attributes.time 0+job-slurm.json | uniq -c | sort -nrk 1
      6 "2025-11-05T07:30:00+01:00"
      5 "2025-11-05T07:45:00+01:00"
      5 "2025-11-05T07:15:00+01:00"
      5 "2025-11-05T05:15:00+01:00"
      5 "2025-11-05T02:30:00+01:00"
      ....
```

PR #438, opened 2025-11-03
Fix #437 - replace gres_detail

### LINKED ISSUE #437: Inconsistent definition of gres_detail
In types.go, GRESDetail is specified as a `[]string`, but without making it clear what the individual strings are.  However, Sonar sends a single string, taken either from ReqTRES or from AllocTRES.  Since Jobanalyzer expects `[]string` it rejects the entire data packet.  https://github.com/NAICNO/Jobanalyzer/issues/818#issuecomment-3480226083.

(Sonar also prefers ReqTRES over AllocTRES, and I'm not sure that is exactly desirable since it means that for a Running job we'll see the request but not the allocation, but that's a separate bug I think, just noting it here for posterity.)

@2maz, note this problem, which will be resolved one way or the other and lead to yet another change somewhere in the ingest path.

I'm not sure yet how to resolve this, need to look at what slurm-monitor did.

It is possible that the right thing to do is to have two fields, request_tres and alloc_tres, and send one or both (or neither) depending on what the data are.  I'm fairly sure that it's the right thing to defer parsing to the client side, but I'm open to being convinced otherwise.

lars-t-hansen:
The important changes here:

- remove the gres_detail field because unclear/underspecified
- introduce requested_resources field which unambiguously comes from scontrol
- introduce allocated_resources field which unambiguously comes from sacct
- fields not present in output if not found in output from those programs, as per normal
- field syntax unambiguously defined by a parser that's present in the Go code alongside the types.go spec

Everything else is just a consequence of those.  The important files are types.go and slurmjobs.rs.


util/formats/newfmt/types.go
```diff
@@ -898,14 +895,17 @@ type SlurmJob struct {
...
 	// scontrol: `ReqTRES`
-	// sacct: `AllocTRES`
+	ReqTRES string `json:"requested_resources,omitempty"`
+
+	// Allocated resources. If present, this comes from sacct's AllocTRES field.  See
+	// DecodeSlurmTRES() in decode_jobs.go in this directory for encoding details.
 	//
-	// slurm: `JOB_INFO.gres_detail`
-	GRESDetail []string `json:"gres_detail,omitempty"`
+	// scontrol: `ReqTRES`
```
lars-t-hansen: Copypasta, this should be "sacct: `AllocTRES`".

PR #446, opened 2025-11-12
Fix #244 - Remove CSV and old-format JSON formatting

### LINKED ISSUE #244: Just say no to CSV
The JSON encoding of sonar ps data on a Fox login node is nearly 40% smaller than the CSV encoding (b/c records do not carry redundant version, timestamp and hostname for every record, and b/c the hostname is fairly long on the login node).  When I looked at these data the other day I got this reversed: I thought the CSV was smaller.

```
[ec-larstha@login-3 sonar-main]$ target/debug/sonar ps --exclude-system-jobs --load | wc
    614     646   94470
[ec-larstha@login-3 sonar-main]$ target/debug/sonar ps --exclude-system-jobs --load --json | wc
      1      33   59042
```

Obviously if we *really* care about data size we'll use some binary encoding or make sure that data are compressed somehow for transmission, ideally very cheaply.

lars-t-hansen:
A lot of the prep work had been done so this is pretty undramatic.


lars-t-hansen:
Not going to rename 'newfmt' to something else, partly because I don't know what that something else should be and partly because this would be disruptive to much of the code and to Jobanalyzer as well.  May consider it in the future when things are clearer.  For now, leave a pointer in util/formats/README.md about why it's the "new" format and where to find information about the old format.

lars-t-hansen:
@bast, gentle review ping.

PR #451, opened 2025-11-18
Fix #450 - compute numa nodes and distances properly

### LINKED ISSUE #450: Sonar fails on VM system due to confusion about NUMA
Joy.  I have an NREC instance of type m1.large which has two vCPUs, running RHEL10.  Here's the important parts of /proc/cpuinfo:
```
[almalinux@sonalyze-backend-test local]$ cat /proc/cpuinfo
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 60
model name      : Intel Core Processor (Haswell, no TSX)
stepping        : 1
microcode       : 0x1
cpu MHz         : 2606.108
cache size      : 16384 KB
physical id     : 0
siblings        : 1
core id         : 0
cpu cores       : 1
apicid          : 0
initial apicid  : 0

processor       : 1
vendor_id       : GenuineIntel
cpu family      : 6
model           : 60
model name      : Intel Core Processor (Haswell, no TSX)
stepping        : 1
microcode       : 0x1
cpu MHz         : 2606.108
cache size      : 16384 KB
physical id     : 1
siblings        : 1
core id         : 0
cpu cores       : 1
apicid          : 1
initial apicid  : 1
```
but something goes wrong somewhere because Sonar errors out:
```
$ sonar sysinfo
{"meta":{"producer":"sonar","version":"0.16.0"},"errors":[{"detail":"Unable to read /sys/devices/system/node/node1/distance","time":"2025-11-17T13:17:21+00:00","node":"sonalyze-backend-test.novalocal"}]}
```
and indeed that isn't there:
```
$ find /sys/devices/system/node -name distance
/sys/devices/system/node/node0/distance
```
and numactl doesn't think it should be either:
```
$ numactl -H
available: 1 nodes (0)
node 0 cpus: 0 1
node 0 size: 7931 MB
node 0 free: 3532 MB
node distances:
node     0
   0:   10
```
so Sonar is being confused about something.  Possibly it is using physical ID instead of core ID?

lars-t-hansen:
OK, so strategy:

- look for the `distance` files for nodes up to numsockets-1
- if we find them, they should together make a square matrix
- and the number of nodes we found is exposed as a new numa_nodes field in the output
- if we don't find them, we do not error out but just return an empty value

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -367,7 +367,7 @@ checksum = "7a2ae44ef20feb57a68b23d846850f861394c2e02dc425a50098ae8c90267589"
 
 [[package]]
 name = "sonar"
-version = "0.16.0"
+version = "0.16.1-devel"
 dependencies = [
  "base64",
  "crossbeam",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sonar"
-version = "0.16.0"
+version = "0.16.1-devel"
 edition = "2021"
 repository = "https://github.com/NordicHPC/sonar"
 
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -14,6 +14,11 @@ relied upon.  Consumers should expect to see the "new" JSON format for new data,
 should ask for that format.
 
 
+## Changes in v0.16.1 (on `release_0_16`)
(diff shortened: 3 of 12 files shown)


2maz (approved):
LGTM

PR #455, opened 2025-11-25
Fix #453 - clean up logging

### LINKED ISSUE #453: Fix the logging situation
At some point, when the supply chain security situation was something we felt we could fix, a simple local logging package was implemented to avoid pulling in something bigger.  But the supply chain security situation became unfixable when we pulled in RdKafka, and RdKafka depends on some logging infra to be able to work at all (it uses the log package), so we should just bite the bullet and lean into this.  (Probably same for command line argument parsing and daemon config file parsing but that's another story.)

Once we have this, the RdKafka logs become available to us, which is important.  And our logging will fit what the Rust ecosystem expects.

For this:
- we should use the standard log crate in our code
- we need to pull in some logging consumer / controller or write our own for that, since the log crate will need that to produce anything

For the second point, I see that the "simple" (their words) env_logger package has "only" 11 dependencies.  It's nice to be able to control things through env vars but a little research is probably warranted on how much complexity we really need for this.

lars-t-hansen:


lars-t-hansen:
Lovely, this takes us to Rust 1.81... :-(

src/datasink/kafka.rs
```diff
@@ -31,7 +30,7 @@ use std::time::{Duration, Instant};
 use crossbeam::channel;
 
 use rdkafka::client::ClientContext;
-use rdkafka::config::ClientConfig;
+use rdkafka::config::{ClientConfig, RDKafkaLogLevel};
```
lars-t-hansen: Undo this, not currently used.  Should run vet.

PR #458, opened 2025-11-27
Fix #457 - grab processes running inside docker

### LINKED ISSUE #457: Grab docker processes
See extensive discussion on #122.  We want to grab - ideally in a lightweight way - processes running inside docker containers.

lars-t-hansen:
This may or may not be done, not sure yet.  It will look for processes called "containerd*" whose ppid is 1, these are the roots of the containers.  Every child of those processes should be included even if --exclude-system-processes is true and the owner is some system uid.  This appears to work for both docker run and docker compose.

Obviously there may be false positives but they are unlikely (it is possible for a user to construct a process meeting the filtering criteria) and mostly benign.  We could consider performing the extra marking if there's at least some docker process running on the system.  Not sure it's worth the extra effort.

Processes in containers get assigned UIDs that are curious, it may be that they are cgroup-relative in some way - I've seen UID 100 and UID 1000 assigned to java processes running inside a container, both were wrong.

### DIFF
--- a/src/ps.rs
+++ b/src/ps.rs
@@ -228,6 +228,15 @@ impl ProcessTable {
 
 pub type ProcInfoTable = HashMap<Pid, ProcInfo>;
 
+#[derive(Clone, Copy, Default, PartialEq)]
+pub enum CState {
+    #[default]
+    Unknown,
+    Root,
+    Child,
+    Not,
+}
+
 #[derive(Clone, Default)]
 pub struct ProcInfo {
     pub user: String,
@@ -237,6 +246,7 @@ pub struct ProcInfo {
     pub rolledup: usize,
     pub num_threads: usize,
     pub is_system_job: bool,
+    pub container_state: CState,
     pub has_children: bool,
     pub job_id: usize,
     pub is_slurm: bool,
@@ -322,6 +332,10 @@ fn collect_sample_data(
         return Ok(None);
     }
 
+    if opts.exclude_system_jobs {
+        procinfo_by_pid = mark_containers(procinfo_by_pid);
+    }
+
     let mut candidates = if opts.rollup {
         rollup_processes(procinfo_by_pid)
     } else {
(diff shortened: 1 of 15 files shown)


lars-t-hansen:
Also this breaks some tests that think that --exclude-system-jobs really mean that, but it doesn't anymore...

Indeed maybe there are some subtleties around what job numbers might mean in this context.

lars-t-hansen:
I think we should expose the is-running-in-a-container flag in the Sonar per-process output.

There will be a "docker" process running somewhere for docker compose and docker run except when run with -d, but the processes running in a container will not be children of that docker process regardless, so it's not clear that this is actionable in any automatic sense.  On the client side, an operator will be able to observe that there is a docker running and will of course see that there are processes running in the containers, and if only one user is running docker may be able to connect the two.

What would be extra neat is if docker records the relationship somewhere eg in cgroup data, I will investigate; we would then expose it as we do the slurm job id.

src/ps.rs
```diff
@@ -228,6 +228,11 @@ impl ProcessTable {
 
 pub type ProcInfoTable = HashMap<Pid, ProcInfo>;
 
+// container_flags has at most one bit set.  If the state is unknown the value is zero.
+pub const CONT_ROOT: i8 = 1;
```
lars-t-hansen: This is not a bit vector, these are exclusive state values, so number 1, 2, 3 and fix the comment.

src/ps.rs
```diff
@@ -349,6 +359,65 @@ fn collect_sample_data(
...
+    // parent, but due to how information is collected there's the chance that a ppid will not exist
+    // in the table, so we must deal with that.
+    //
+    // The separate collecting of keys sucks but is a fact of Rust.
+    let all = procinfo_by_pid.keys().map(|v| *v).collect::<Vec<Pid>>();
+    let mut keys = vec![];
+    for k in all {
+        keys.clear();
+        let mut current = k;
+        let mut flagbits = 0i8;
```
lars-t-hansen: These are not actually "flag bits", it's a state probably.

src/ps.rs
```diff
@@ -228,6 +228,11 @@ impl ProcessTable {
 
 pub type ProcInfoTable = HashMap<Pid, ProcInfo>;
 
+// container_state has one of these values, or is zero if the state is as-yet unknown.
+pub const CONT_ROOT: i8 = 1;
```
lars-t-hansen: I guess this could be an enum now.

src/ps.rs
```diff
@@ -228,6 +228,16 @@ impl ProcessTable {
 
 pub type ProcInfoTable = HashMap<Pid, ProcInfo>;
 
+// container_state has one of these values, or is zero if the state is as-yet unknown.
```
lars-t-hansen: Groan, outdated comment.

src/ps.rs
```diff
@@ -349,6 +364,65 @@ fn collect_sample_data(
...
+        let mut state = CState::Unknown;
+        while let Some(v) = procinfo_by_pid.get(&current) {
+            state = v.container_state;
+            if state != CState::Unknown {
+                break;
+            }
+            keys.push(current);
+            current = v.ppid;
+        }
+        // Invariant: every node in keys has state zero and unless the search terminated in a
```
lars-t-hansen: Outdated comment.

PR #469, opened 2025-12-17
Fix #468 - Send cluster msg on startup

### LINKED ISSUE #468: Send cluster message at sonar startup if configured
This is like sysinfo being sent at startup - it is useful for the backend to have this data as soon as possible.  Probably there should be an option just as for sysinfo, default on.

lars-t-hansen:


tests/daemon-startup.sh
```diff
@@ -0,0 +1,55 @@
...
+output=$(tmpfile daemon-startup-output)
+log=$(tmpfile daemon-startup-log)
+inifile=$(tmpfile daemon-startup-ini)
+
+# The startup message will fire (should fire) and then sonar will exit.  If they don't fire, sonar
+# will hang because the cadence is very, very slow (8 days).  For that reason, kill it if it is
+# still there after a few seconds.
+#
+# With very low probability, the timed messages will fire at the same time as the startup message.
+# Over time, this probability approaches zero.
```
lars-t-hansen: That's a clunky way of saying that the test may intermittently fail but it will be obvious that it is an intermittent.  Should probably reword.

PR #473, opened 2026-01-21
Fix #445 - Capture disk stats

### LINKED ISSUE #445: Take some readings from vmstat -D
So vmstat -D:
```
$ vmstat -D
            3 disks
            3 partitions
       173461 total reads
        20943 merged reads
     16081962 read sectors
       101573 milli reading
       370429 writes
        72961 merged writes
     19979722 written sectors
      3137102 milli writing
            0 inprogress IO
          104 milli spent IO
         3274 milli weighted IO
```
This is per-node and not per-process but on some clusters this will be actionable.  Should look at the vmstat source to find out where these come from and maybe include them in the system profiling data.

It doesn't look like vmstat is suid root so hopefully the data are available to us, TBD.  UTSL.

lars-t-hansen:
Capture the values of /proc/diskstats to at least give us a basic idea of what happens on the various disks.  This is very linux-specific, and we could instead make a selection of fields with plausible os-independent semantics, but I'm not really sure it's worth the trouble.  Presumably other OS's stats could be mapped onto the fields defined by Linux, for the most part, if it comes to that.


2maz (approved):
LGTM

PR #484, opened 2026-01-30
Fix #483 - Synthesize PIDs for rolled-up processes

### LINKED ISSUE #483: Assign synthetic PID to rolled-up process tree in daemon mode
This will require keeping some side table for the job that hashes the rollup criteria (command line, ppid, ...) into a synthetic PID that could not be a true PID.  the rolledup count will be sufficient to flag that the PID is synthetic, but even so the new PID should be outside the valid Linux PID range (typically 2^22 but we can query the system somehow, anyway we have a uint64 range to play with).

It's sort of a bug that we didn't do this in the first place but it requires state in Sonar to deal with the shifting sets of processes in a rolled-up job so it only works in daemon mode.  The spec will be changed to note these complications.

It will require some kind of garbage collection of the synthetic PID table, which is a pain.

lars-t-hansen:


src/ps.rs
```diff
@@ -38,10 +38,237 @@ pub struct PsOptions {
     pub cpu_util: bool,
 }
 
+// The PidMap is used for rolled-up pid synthesis.  A rolled-up process - a (job, parent, command)
+// triple - is entered into the map when it is first encountered and holds the synthesized pid of
+// that process.  Whenever it is encountered it is marked dirty.  At the end of processing a full
+// set of samples, the map can be scanned and non-dirty elements can be removed.  If the pool of
+// available pids is getting low, a scan of the pid space can be made to find more at that time.
```
lars-t-hansen: This is different now, and indeed it's not clear why we would sweep at this time.

src/ps.rs
```diff
@@ -38,10 +38,237 @@ pub struct PsOptions {
...
+        }
+    }
+
+    fn maybe_collect(&mut self, system: &dyn systemapi::SystemAPI) {
+        let limits = get_limits(system);
+
+        self.tick += 1;
+        if self.tick >= limits.collection_interval {
+            self.mark();
+            self.sweep(system);
```
lars-t-hansen: We don't need to sweep here, sweeping happens on demand.  And then the mark function can be inlined I think.

lars-t-hansen:
@2maz FYI.  I'm reasonably happy with this so I'll merge it, and then try to deploy to Fox first thing Monday, if I can't think of any red flags over the weekend.

PR #485, opened 2026-02-02
For #483 - Refactor to make it better

lars-t-hansen:
Clean up the PidMap.


src/pidmap.rs
```diff
@@ -0,0 +1,274 @@
...
+
+#[derive(Eq, Hash, PartialEq)]
+struct ProcessKey {
+    job_id: JobID,
+    ppid: Pid,
+    command: String,
+}
+
+struct ProcessValue {
+    pid: u64,
```
lars-t-hansen: Should probably be Pid.

PR #506, opened 2026-02-17
Fix #493 - assets for building RPMs

### LINKED ISSUE #493: Build Sonar RPM for Saga
I think:

- an RPM with a binary but no source
- x86_64 for Saga only
- compiled on Saga with local rustc
- ideally we can use the existing compiled gpu support blobs (are there gpus on saga?  if not, should we compile w/o gpu support?)
- we'll branch for 0.18.0 and base it off of that
- systemd files all go into it
- it goes into /var/lib/sonar on each node
- some of the sigma2 requirements on systemd files TBD

lars-t-hansen:
This is not actually for Saga but is a generic RPM build setup, with a couple of example cases and assets that can go into the RPM, and instructions.  The Saga-specific code will be in the sonar-deploy repo on NRIS gitlab.

### DIFF
--- a/build-dist/rpm-assets/sonar.cfg
+++ b/build-dist/rpm-assets/sonar.cfg
@@ -1,6 +1,6 @@
 # Sonar daemon stub config file, with no operations and no data exfiltration.
 #
-# For how to configure and use sonar, see https://github.com/NordicHPC/sonar/doc/MANUAL.md.
+# For how to configure and use Sonar, see https://github.com/NordicHPC/sonar/doc/MANUAL.md.
 #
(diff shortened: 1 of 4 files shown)


build-dist/rpm-assets/sonar.cfg
```diff
@@ -0,0 +1,75 @@
+# Sonar daemon stub config file, with no operations and no data exfiltration.
+#
+# For how to configure and use sonar, see https://github.com/NordicHPC/sonar/doc/MANUAL.md.
```
lars-t-hansen: "Sonar"

PR #507, opened 2026-02-17
Fix #505 - better logging from kprox

### LINKED ISSUE #505: Kafka HTTP proxy logging
The logging situation needs to be improved for production:

* With -v, all verbose logging should go to the journal.
* With -d, all debug logging should go to stderr (it already does) and all error/verbose logging should go to stderr in addition to the journal.

Possibly, if we were to standardize on something like the slurm-monitor ingestion UI for user-facing logging, we could add an option to send information to that UI via zeromq, see https://zeromq.org/languages/go/ for available (trusted) libraries.

lars-t-hansen:
Tested manually locally.  With -d we get logging both in the journal and on stderr, with only -v we get it only in the journal, without any options we only get critical messages in the journal.

### DIFF
--- a/util/kafka-proxy/kprox.go
+++ b/util/kafka-proxy/kprox.go
@@ -156,7 +156,7 @@ type Msg struct {
 
 func main() {
 	var err error
-	syslogger, err = syslog.Dial("", "", syslog.LOG_INFO|syslog.LOG_USER, "kprox")
+	syslogger, err = syslog.Dial("", "", syslog.LOG_USER, "kprox")
 	if err != nil {
 		log.Fatalf("Failing to open syslogger: %v", err)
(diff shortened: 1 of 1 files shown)


util/kafka-proxy/kprox.go
```diff
@@ -398,29 +408,22 @@ func parsePayload(ch chan<- Msg, payload []byte) (int, string) {
 		}
 		loc := bytes.IndexByte(payload[ix:], '\n')
 		if loc == -1 {
-			if *verbose {
-				log.Printf("Trailing junk in message")
-			}
+			log.Print("Trailing junk in message")
```
lars-t-hansen: Bug.

PR #517, opened 2026-02-20
Fix #516 - hostname pruning

### LINKED ISSUE #516: Introduce host name pruning
There are issues with how host names are reported in data, see #459 for a longer discussion, but it stems from how nodes are configured, how slurm is configured, and how the back-end interprets the data.  Various hacks have been attempted but the principled solution that will work on many clusters is to strip host names down to their leaf element in Sonar and only report this.  This plays well with Slurm data and is easy to understand and implement.  Back-ends become simpler too (sonalyze has a significant amount of logic to deal with various ways of expressing host names).

For 0.18 I think we do this, which is not very invasive and is easily tested:

- Spec work
  - Restate the type for Cluster as `FQDN`
  - Rename the type NodeRange as HostnameRange
  - Clarify in prose that the expanded values in HostnameRange are of type Hostname
  - For the Hostname type, clarify that it can be modified in various ways depending on Sonar flags
- Introduce a setting global.hostname-only : bool, default false (for compatibility), if this is set then all Hostname values in the output are stripped to their leaf component, ie, c1-10.fox becomes c1-10, int-1.fox.ad.fp.educloud.no becomes int-1.
- Make this setting a command line flag, also
- Make it illegal (for now) to use cluster.domain with global.hostname-only
- Do not introduce jobs.domain at this stage
- Document it all

lars-t-hansen:
Noteworthy for this patch is that external code that depends on the type definitions - sonalyze and slurm-monitor both - may need to be adapted slightly.  No representations have changed but some of the names have changed, and the interpretations have been narrowed.

### DIFF
--- a/build-dist/rpm-assets/sonar.cfg
+++ b/build-dist/rpm-assets/sonar.cfg
@@ -11,6 +11,7 @@ cluster = invalid-cluster.no
 role = node
 # lock-directory =
 # topic-prefix =
+# hostname-only = false
 
 # Output is to a a Kafka broker (directly or via a proxy over HTTP(S)), a local directory tree, or
 # to stdout.  These data sinks are mutually exclusive.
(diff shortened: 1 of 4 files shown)


lars-t-hansen:
Oops, need to update documentation also.

PR #533, opened 2026-03-10
update CITATION.cff; closes #529

### LINKED ISSUE #529: Update CITATION.cff
This still cites 0.6.2, should it cite the latest release?

This still has Radovan's old email address, should it have a new one (or mine, or something else)?

Should update both on 0.18 and main.

bast:
Once this is reviewed/merged, I will send a similar PR to the 0.18 branch.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -8,8 +8,6 @@ type: software
 authors:
   - given-names: Radovan
     family-names: Bast
-    email: radovan.bast@uit.no
-    affiliation: UiT The Arctic University of Norway
     orcid: 'https://orcid.org/0000-0003-4498-3806'
   - given-names: Mathias
     family-names: Bockwoldt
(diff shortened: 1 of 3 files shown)


bast:
@lars-t-hansen I have modified build-dist/version.go to make it part of "make generate". Please let me know if you would prefer this to be done in a different place.

lars-t-hansen:
@bast I think this is close but not quite what I would want.  build-dist/ is for RPM and DEB builds so let's not confuse it with other things.  What I would do here is change the generate: target in the top-level Makefile to depend on CITATION.cff as you have and then for the latter to depend on Cargo.toml, ditto; then have a new citation-version.go (or .py if you prefer) at the top level to generate that.  If you think that is too messy, create util/citation/ and put the version.go in there maybe.

bast:
Thanks for feedback! I have pushed an update - hopefully it is closer now?

lars-t-hansen:
LGTM.

bast:
Thanks! Now marked as ready to merge.

Issue #540, opened 2026-04-07
Make an http-only exfiltration path


lars-t-hansen:
Simplification:

For the Sigma2 systems, and I suspect in general, an http-only path is really what we want, not Kafka.  Probably data gets sent to a host under a base API url, which could be configurable but here we'll say is "/add", then the POST URL is "/add/CLUSTER/DATATYPE/TIMESTAMP", and the auth is probably in the headers somehow, it may depend on the auth scheme (for HTTP simple auth it's in the header, for tokens it probably also is in the header).  It is advantageous if the data don't have to be parsed to authenticate but can just be stored.  Auth should be optional.

There would be an `[http]` data sink section, precluding the others, and the options would be similar to the ones in the Kafka section.

It would be possible to build Sonar without Kafka (reducing size and dependencies).

This path could start using Curl as we do for Kafka but ideally we move to reqwest eventually, see #525.

PR #541, opened 2026-04-14
For #540 - Refactor http code for reuse

lars-t-hansen:
This splits a bunch of logic out of the Kafka code into a "background uploader" idea and a "http post" idea.  Functionality-wise this should represent no change.  I will need to test some more before landing but otherwise this is done.  Once it's done, #540 is easy.

### DIFF
--- a/src/datasink/background.rs
+++ b/src/datasink/background.rs
@@ -1,0 +1,79 @@
+use crate::util::rng::Rng;
+use crossbeam::channel;
+use std::thread;
+use std::time::{Duration, Instant};
+
+pub struct Msg {
+    pub timestamp: u64,
+    pub topic: String,
+    pub key: String,
+    pub value: String,
+}
+
+impl Msg {
+    // This can return an estimate, but it's better to overestimate than underestimate.
+    pub fn size(&self) -> usize {
+        return self.topic.len() + self.key.len() + self.value.len() + 20;
+    }
+}
+
+pub enum Message {
+    Stop,
+    M(Msg),
+}
+
+pub trait BackgroundSender {
+    fn send_all(&self, id: usize, msgs: Vec<Msg>);
+    fn shutdown_delay_ms(&self) -> usize;
+}
+
+pub fn background_producer(
+    sending_window: u64,
+    incoming_message_queue: channel::Receiver<Message>,
+    sender: &dyn BackgroundSender,
+) {
+    let mut id = 0usize;
+    let mut rng = Rng::new();
+    let mut timeout: channel::Receiver<Instant> = channel::never();
+    let mut armed = false;
+    let mut must_arm = false;
+    let mut backlog = Vec::new();
+
+    'producer_loop: loop {
+        if must_arm {
+            assert!(!armed);
+            let sleep = rng.next() as u64 % sending_window;
+            // Note, the /Sleeping {} before sending/ pattern is used by regression tests.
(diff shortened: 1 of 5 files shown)


src/datasink/background.rs
```diff
@@ -0,0 +1,139 @@
...
+                    }
+                    sz = newsz;
+                    i += 1;
+                }
+                if i == 0 {
+                    log::error!(
+                        "Message of size {} is too large to send, should not happen",
+                        backlog[0].size()
+                    );
+                    // TODO: Must drop it or send it!!
```
lars-t-hansen: Oops, fix this.

src/datasink/background.rs
```diff
@@ -0,0 +1,139 @@
...
+fn send_all(
+    sender: &dyn BackgroundSender,
+    cutoff: Option<usize>,
+    mut id: usize,
+    mut backlog: Vec<Msg>,
+) {
+    if !backlog.is_empty() {
+        // Note, the /Sending {} items/ pattern is used by regression tests.
+        log::debug!("Sending {} items", backlog.len());
+        let (boverhead, moverhead) = sender.metadata_size();
```
lars-t-hansen: How about bOverhead and mOverhead

src/datasink/background.rs
```diff
@@ -0,0 +1,139 @@
...
+                Ok(Message::Stop) | Err(_) => {
+                    send_all(sender, cutoff, id, backlog);
+                    break 'producer_loop;
+                }
+            }
+        }
+    }
+
+    // Best effort: give the sending thread an opportunity to send what it has.
+    thread::sleep(Duration::from_millis(2 * sender.shutdown_delay_ms() as u64));
```
lars-t-hansen: Not clear why 2*x here instead of in the methods, that would be cleaner.

src/datasink/kafka.rs
```diff
@@ -227,76 +210,51 @@ fn kafka_producer(
...
                 }
             }
         }
     }
 
-    // Best effort: give the Kafka thread an opportunity to send what it has.
-    thread::sleep(Duration::from_millis(2 * KAFKA_BUFFER_MS as u64));
-}
+    fn shutdown_delay_ms(&self) -> usize {
+        KAFKA_BUFFER_MS
```
lars-t-hansen: Times 2

src/datasink/kafka.rs
```diff
@@ -422,289 +380,85 @@ fn make_sender_adapter(
...
-        }
+    }
+
+    fn shutdown_delay_ms(&self) -> usize {
+        5000
+    }
+
+    fn metadata_size(&self) -> (usize, usize) {
+        // Conservative overhead for punctuation, field names, etc of the control object
+        (10 /* per batch */, 200 /* per message */)
```
lars-t-hansen: topic and key are accounted for in the message size actually, so the 200 is probably generous.

lars-t-hansen:
Despite --silent and --retry, curl spams a lot on stderr when it can't contact the server.  Sonar handles this but not very gracefully, it will dump a debug message every time it can read something, it does not wait for a complete error message, so sometimes there can be lots of debug spew for what's one line from curl.  This is more or less by design since not all programs write complete lines but since it's curl I guess we could assume there will be a newline eventually.  Since the problem is not new I'll file a followup, not fix it in this patch.

PR #542, opened 2026-04-20
Fix #540 - http data sink

### LINKED ISSUE #540: Make an http-only exfiltration path

lars-t-hansen:


doc/HOWTO-DAEMON.md
```diff
@@ -108,6 +108,34 @@ The `ca-file`, `sasl-password` and `sasl-password-file` are explained in
...
+http-proxy = <url>
+http-payload-limit = <volume value>             # default none
+sending-window = <duration value>               # default 5m
+timeout = <duration value>                      # default 30m
+upload-password = <string>                      # default none
+upload-password-file = <string>                 # default none
+```
+
+The `api-root` is the root of the upload URL, see "HTTP sink" below.  For the Sonalyze v0 API it
+would normallly be `https://<sonalyze-server-and-port>/api/v0`.
```
lars-t-hansen: "lll"

lars-t-hansen: Also, the URL is wrong.  For an upload URL for v0 the path root would be /api/v0/insert.

doc/HOWTO-DAEMON.md
```diff
@@ -108,6 +108,34 @@ The `ca-file`, `sasl-password` and `sasl-password-file` are explained in
...
+
+The `http-proxy`, `http-payload-limit`, `sending-window`, and `timeout` are as for Kafka, above.
+
+The `upload-password` and `upload-password-file` are strings, mutually exclusive.  These will be
```
lars-t-hansen: It should be "contents of the `upload-password-file`"

doc/MANUAL.md
```diff
@@ -9,15 +9,18 @@ cluster status.  It is always running and can be used for post-hoc and on-line a
...
 
 Alternatively, data can be stored on-node in a directory tree; this is mostly useful for single-node
 non-Slurm configurations, as in some cloud deployments.
 
 Sonar operation is driven by a configuration file that determines how and how often to collect and
 exfiltrate the data.
 
+[Companion tools](#companion-tools) exist for data access and analysis.
+
+Sonar will need to be allowed to create (small, short-lived) temp files on the local file system.
```
lars-t-hansen: ... ikn the standard temp directory.

src/datasink/http.rs
```diff
@@ -0,0 +1,179 @@
...
+    fn sending_window_s(&self) -> u64 {
+        self.settings.sending_window.to_seconds()
+    }
+
+    fn shutdown_delay_ms(&self) -> u64 {
+        5000
+    }
+
+    fn batch_size(&self) -> Option<usize> {
+        self.settings.http_payload_limit
```
lars-t-hansen: This needs to be removed because HTTP messages are currently always sent singly - there is no batching at present.  We should remove the option and always return 0 here.  The message size can still return the true size.

src/datasink/http.rs
```diff
@@ -0,0 +1,179 @@
+// The HttpSink provides exfiltration over HTTP POST in a native way, not to the Kafka proxy.
+//
+// This shares a lot of structure with the Kafka sink, especially the Kafka-over-HTTP proxy
+// part, but this is so simple I've not bothered to factor common parts.
```
lars-t-hansen: Remove this remark, it is no longer true.

src/datasink/http.rs
```diff
@@ -0,0 +1,179 @@
...
+    fn stop(&mut self, _system: &dyn SystemAPI) {
+        let _ = self.outgoing_message_queue.send(Message::Stop);
+        let mut producer = None;
+        std::mem::swap(&mut producer, &mut self.producer);
+        let _ = producer.unwrap().join();
+    }
+}
+
+fn http_producer(
+    curl_cmd: &str,
```
lars-t-hansen: This could just take a String.

PR #548, opened 2026-05-18
No bug: Update crates

lars-t-hansen:
Pull new minor versions across the board with cargo update + bump some pinned versions.

### DIFF
--- a/.github/workflows/test-minimal.yml
+++ b/.github/workflows/test-minimal.yml
@@ -23,8 +23,8 @@ jobs:
       run: |
         curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
         source $HOME/.cargo/env
-        rustup install 1.81.0
-        rustup default 1.81.0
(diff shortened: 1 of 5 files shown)


lars-t-hansen:
Well, as usual this may require bumping the Rust version - this time to the 2024 edition, starting with 1.85.  Given additional dependencies that means at least 1.88.0 (due to `time` which is required by `rdkafka` I think) which was released in June 2025.  1.88.0 is the newest version available on Fox and Saga so that's OK (of course a personal install is always possible, but still).  My account on eX3 appears to have been closed so I can't test what's available there.  On Olivia the module system seems to be working in some other way and I'm not sure what's there.

lars-t-hansen:
We're going to have to move to the 2024 ed at some point so might as well now.  But in that case, I'll probably want to roll the update of subprocess from current prerelease to 1.1.0 into this patch.  That has an API change, so it may take a little more work.

lars-t-hansen:
Not going to update subprocess here, will do that as separate work item.

PR #561, opened 2026-05-21
Fix #559 - let kprox perform authorization

### LINKED ISSUE #559: kprox should be able to authenticate/authorize uploads
To simplify the backend setup, we could add simple functionality to the kafka-proxy so that it is it, and not Kafka, that authenticates the uploader / authorizes the upload.  If we have that, and the kafka-proxy runs on the same host as the Kafka broker and slurm-monitor and sonalyze (as will normally be the case), then the back-end need only have a stock Kafka instance running out of a standard docker with no config changes, and neither slurm-monitor nor sonalyze nor kafka-proxy need be configured for crypto when talking to the broker.

This would take the form of a property in the `[http]` section of the kafka-proxy ini, which would point to a file containing clustername:password pairs.  These would be matched against the incoming sasl credentials in the proxied message.  The `kafka.sasl` setting would be false in this case.  The presence of the password file name in the http section would imply authorization checking on ingress.

This will put a little bit more pressure on #479 but not really enough to worry.

lars-t-hansen:


doc/HOWTO-KAFKA.md
```diff
@@ -253,9 +253,19 @@ that file, along with a `kafka.rest-endpoint` using HTTPS (as above).  On the pr
...
 
+### Authorizing the upload
+
+The upload can be authorized either by the the kafka-proxy forwarding the SASL credentials that are
+in the message to Kafka, and Kafka carrying authorization information, or by the kafka-proxy itself
+performing the authorization and not forwarding those fields to Kafka.  The advantage of the latter
+is a simpler Kafka configuration.
+
+In the latter case, create an upload identity file with `cluster-name:upload-password` lines and set
+the `http.upload-auth-file` to point to that file.
```
lars-t-hansen, line 264: "property"

Issue #508, opened 2026-02-17
Intermittent failure: daemon-rollup2


lars-t-hansen:
This test sometimes fails sometimes on my workstation.  The PID GC mechanism runs out of PIDS.  The test uses a small pool for the test but if there are process trees - esp browser processes - on the system, we may be sunk.

PR #563, opened 2026-05-27
Fix #562 - Use 2024 Edition

### LINKED ISSUE #562: Update to 2024 edition
Recently we pulled crates that require increasing MSRV to Rust 1.88.0, which is the 2024 Edition (starting with 1.85).  We should change the edition in Cargo.toml to 2024 as well, in principle it's good to keep up.  But this will require some tweaks to the code, as some things have changed in the language: extern blocks (in the C interface) must be marked unsafe, and there are some uses of `if let Some(ref ...) ...` that cause the compiler to complain, probably left over in fairly old code.

lars-t-hansen:
This currently fails tests/daemon-rollup2 with a panic in the pidmap.  That test has always been a little sketchy but this is not expected at all.

Other than that, this was straightforward.  The 2024 edition has some formatting tweaks so this patch has some formatting changes because of that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "sonar"
 version = "0.19.0-devel"
-edition = "2021"
+edition = "2024"
 repository = "https://github.com/NordicHPC/sonar"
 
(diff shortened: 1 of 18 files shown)


lars-t-hansen:
The test failure is just #508, again.

Issue #524, opened 2026-02-24
Use bindgen to generate Rust from GPU shims


lars-t-hansen:
We do not currently use [bindgen](https://github.com/rust-lang/rust-bindgen) to generate the Rust/C interface to the GPU shims, but we should.  (The interfaces are manually maintained, which was fine when there were two GPU types but now there are four plus the "fake" GPU and there's some room for error in doing this manually, in addition to the tedium.)  This change will introduce a developer dependency on `clang-devel` and maybe `bindgen-cli` - not the end of the world hopefully.

See eg https://rust-lang.github.io/rust-bindgen/command-line-usage.html.  What's emitted using that template is basically right but has two problems: there's a bunch of gunk from inttypes.h being included in the header file (using stdint.h is a little better), but worse, we need to find some way to insert the linking metadata that's currently in the hand-written versions of this.

PR #569, opened 2026-09-15
Fix #524 - Use bindgen for interface to GPU SMI code

lars-t-hansen:


lars-t-hansen:
Need to test this on at least one system with a GPU...

lars-t-hansen:
Works on ml7 (nvidia) and ml4 (amd).  I no longer have access to Habana and XPU.

Issue #552, opened 2026-05-18
Update subprocess crate


lars-t-hansen:
We're on the prerelease branch still, and that's dead.  We should move to 1.1.0 (as of writing) to avoid being left behind on future bug fixes, but that branch has API changes so it's a little more work with coding and testing.

PR #570, opened 2026-09-15
Fix #552 - Update subprocess crate

lars-t-hansen:
Update the subprocess crate to 1.2.1.

This is probably good to go but I want to run tests on a system that has scontrol / sinfo / sacct before landing.


lars-t-hansen:
Works OK on a Fox login node.

